PR #26, opened 2022-01-31
Handle matching on any valid user object

laurawarr:
- Update parsers to have `orderedParameterPatterns` and `namedParameterPatternMap`. One or both of these are used to build a pattern to match ordered and/or named parameters
- Update user capture pattern for each language parser

### DIFF
--- a/src/utils/diff/parsers/common.ts
+++ b/src/utils/diff/parsers/common.ts
@@ -60,10 +60,12 @@ export abstract class BaseParser {
     abstract identity: string
     matchClientName = true
     clientNames: string[]
+
     abstract variableMethodPattern: RegExp
-    variableMethodKeywordPattern : RegExp | null = null
-    abstract variableNameCapturePattern: RegExp
-    abstract defaultValueCapturePattern: RegExp
+    orderedParameterPatterns: RegExp[] | null = null
+    namedParameterDelimiter = '='
+    namedParameterPatternMap: Record<string, RegExp> | null = null
+
     commentCharacters: string[] = ['//']
 
     constructor(extension: string, protected options: ParseOptions) {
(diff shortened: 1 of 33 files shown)


src/utils/diff/parse.ts
```diff
@@ -4,7 +4,12 @@ import { AndroidParser, CsharpParser, GolangParser, IosParser, JavaParser,
 import { ParseOptions, VariableMatch } from './parsers/types'
 import { CustomParser } from './parsers/custom'
 
-const PARSERS: Record<string, (typeof NodeParser)[]> = {
+type Parser = typeof CustomParser | typeof NodeParser | typeof ReactParser | typeof JavascriptParser
```
ajwootto, line 7: do you need to do this? from the perspective of this file the parsers should all look the same (which is why I just chose one to represent the type)

src/utils/diff/parsers/common.ts
```diff
@@ -60,27 +60,48 @@ export abstract class BaseParser {
...
-            `(?:${this.variableMethodKeywordPattern.source}|${this.variableMethodPattern.source})` :
-            this.variableMethodPattern.source
-        const clientNamePattern = this.matchClientName ? new RegExp(`(?:${this.clientNames.join('|')})`).source : ''
+        const clientNamePattern = this.matchClientName
+            ? new RegExp(`(?:${this.clientNames.join('|')})`).source
+            : ''
+        const orderedParameters = this.orderedParameterPatterns
+            ? this.orderedParameterPatterns
+                .map((p) => p.source)
+                .join(new RegExp('\\s*,\\s*').source)
```
ajwootto, line 82: you could use a Regex literal here to avoid double escaping the backslashes 
`/\s*,\s*/`

src/utils/diff/parsers/common.ts
```diff
@@ -128,24 +149,27 @@ export abstract class BaseParser {
...
         while (cursorPosition < content.length) {
             const substring = content.slice(cursorPosition)
-            const match = this.match(substring, options)
+            const match = this.match(substring)
 
             if (!match) break
 
-            const [matchContent, variableName] = match
+            const [matchContent] = match
+            const variableName = match[1] || match[2] // position or named parameter match
```
ajwootto, line 172: how come it's a different position in the capture groups depending on the pattern? Also would match[1] be falsey while match[2] is truthy?

src/utils/diff/parsers/golang/index.ts
```diff
@@ -1,13 +1,20 @@
 import { BaseParser } from '../common'
 
+const authCapturePattern = /\w*/
```
ajwootto, line 3: minor thing but I've been using the word "Capture" in the variable name to denote that its meant to capture something with a capture group, vs. a regular "pattern"

test/commands/diff.test.ts
```diff
@@ -20,6 +20,12 @@ DevCycle Variable Changes:
 	   Location: test/utils/diff/sampleDiff.js:L10
   4. multi-line-comment
 	   Location: test/utils/diff/sampleDiff.js:L20
+  5. user-object
```
ajwootto, line 23: perhaps we should have this use a separate diff then the more detailed unit testing of the diffing logic? These tests are mainly meant to test the e2e, as long as a valid diff produces a valid comment I don't think we need to test every one of the matching cases in here

src/utils/diff/parse.ts
laurawarr: Since some of the properties are optional (for example orderedParameterPatterns), `typeof IosParser` and `typeof NodeParser` are different

ajwootto: shouldn't those be "protected" then? The actual external signature of all of these should be the same

src/utils/diff/parsers/common.ts
laurawarr: If a language supports both ordered parameters and named parameters, then the resulting pattern will have two capture groups. Only one of the two will match and the other will be undefined

src/utils/diff/parsers/golang/index.ts
laurawarr: True, do you think all the others should be updated that have non-capturing groups?

test/commands/diff.test.ts
laurawarr: True, I'll update these to use an e2e diff rather than the node diff

src/utils/diff/parsers/golang/index.ts
ajwootto: depends how useful you find that distinction in the variable names..it seems like it might be a bit of effort to change for not much gain

ajwootto (approved):

PR #29, opened 2022-02-02
DVC-2089 detect unknown variables

ajwootto:
Add detection for variables that are "unknown" ie. are themselves variables, and not plain strings.

Collect these cases and output them as "notices" in the diff command, alerting the user that they should add an alias in order to detect the variable.

### DIFF
--- a/src/commands/diff/index.ts
+++ b/src/commands/diff/index.ts
@@ -18,7 +18,10 @@ const EMOJI = {
 
 type MatchesByType = {
     add: Record<string, VariableMatch[]>,
-    remove: Record<string, VariableMatch[]>
+    remove: Record<string, VariableMatch[]>,
+    addUnknown: Record<string, VariableMatch[]>,
+    removeUnknown: Record<string, VariableMatch[]>
+
 }
 
 type MatchEnriched = {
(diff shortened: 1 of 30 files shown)


src/utils/diff/parsers/types.ts
```diff
@@ -1,11 +1,14 @@
+export type MatchKind = 'regular' | 'unknown'
 export type VariableMatch = {
     name: string,
     line: number,
     fileName: string,
-    mode: 'add' | 'remove'
+    mode: 'add' | 'remove',
+    kind: MatchKind
```
laurawarr, line 7: It isn't obvious what `kind` means, what about having a prop `unidentified` that's true for unidentified matches?

src/commands/diff/index.ts
```diff
@@ -161,13 +171,37 @@ export default class Diff extends Base {
...
 
+        const enrichedAddUnknown: Record<string, MatchEnriched> = {}
+        const enrichedRemoveUnknown: Record<string, MatchEnriched> = {}
+
+        for (const key of Object.keys(matchesByType.addUnknown)) {
+            enrichedAddUnknown[key] = {
+                variable: null,
+                matches: matchesByType.addUnknown[key]
+            }
+        }
```
laurawarr, line 182:
```suggestion
        for (const [key, matches] of Object.entries(matchesByType.addUnknown)) {
            enrichedAddUnknown[key] = {
                variable: null,
                matches
            }
        }
```

src/commands/diff/index.ts
```diff
@@ -225,9 +267,15 @@ export default class Diff extends Base {
     ) {
         Object.entries(matchesByVariable).forEach(([variableName, enriched], idx) => {
             const matches = enriched.matches
-            const hasNotice = this.useApi() && !enriched.variable
+            const hasNotice = mode === 'add' && ((this.useApi() && !enriched.variable)
+                || enriched.matches.some((match) => match.kind === 'unknown'))
```
laurawarr, line 271:
```suggestion
            const notFound = this.useApi() && !enriched.variable
            const isUnknown = enriched.matches.some((match) => match.kind === 'unknown'))
            const hasNotice = mode === 'add' && (notFound || isUnknown)
```

PR #31, opened 2022-02-02
Detect variables that are not plain strings

ajwootto:
Add logic for detecting when a variable is not a plain string, and output such cases as "notices" for the user to add an alias

### DIFF
--- a/src/utils/diff/parsers/common.ts
+++ b/src/utils/diff/parsers/common.ts
@@ -1,5 +1,5 @@
 import parse from 'parse-diff'
-import { ParseOptions, VariableMatch } from './types'
+import { MatchKind, ParseOptions, VariableMatch } from './types'
 
 type Range = {
     start: number
@@ -16,6 +16,7 @@ type Match = {
     name: string
     // Start/end indicies of matching substring
     range: Range
+    kind: MatchKind
 }
 
 class ParsedChange {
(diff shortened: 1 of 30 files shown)


package.json
```diff
@@ -70,7 +70,7 @@
     "postpack": "shx rm -f oclif.manifest.json",
     "posttest": "yarn lint",
     "prepack": "yarn build && oclif manifest && oclif readme",
-    "test": "mocha --forbid-only \"test/**/*.test.ts\"",
+    "test": "mocha --forbid-only \"test/**/nodejs.test.ts\"",
```
laurawarr, line 73: I'm assuming this is intended to be temporary?

laurawarr: you can run `yarn test -g nodejs`

ajwootto: nice I didn't know that

src/utils/diff/parsers/common.ts
```diff
@@ -149,17 +184,30 @@ export abstract class BaseParser {
...
+    match(content: string): MatchResult | null {
+        const pattern = this.buildRegexPattern()
+
+        const matches = pattern.exec(content)
+        if (matches) {
+            const varName = matches[1] || matches[2]
+            return {
+                isUnknown: !varName.match(/^["'].*["']$/),
+                // position or named parameter match
+                name: varName.replace(/["']/g, ''),
```
laurawarr, line 196:
```suggestion
                name: varName.replace(/^['"]|['"]$/g, ''),
```
what about something like this? to only replace quotes at the start or end of the string

src/utils/diff/parsers/common.ts
```diff
@@ -149,17 +184,30 @@ export abstract class BaseParser {
...
+        const pattern = this.buildRegexPattern()
+
+        const matches = pattern.exec(content)
+        if (matches) {
+            const varName = matches[1] || matches[2]
+            return {
+                isUnknown: !varName.match(/^["'].*["']$/),
+                // position or named parameter match
+                name: varName.replace(/^["']|["']$/g, ''),
+                matches: matches
```
laurawarr, line 197: we could just add the content and index directly to this object, rather than the entire RegExpExecArray

laurawarr (approved):
looks like there's one warning about an unused `AxiosError` in `api/variables.ts` that could be cleaned up

PR #32, opened 2022-02-02
Accept variable aliases

laurawarr:
- Add `--var-alias` flag that accepts multiple variable aliases
- Add `variableAliases` to config file
- If a variable isUnknown, check if an alias is defined
- If an alias exists:
    - Unset `isUnknown`
    - Update `name` property
    - Set `alias` property
- Split config file types into their own file


src/commands/diff/index.ts
```diff
@@ -117,6 +137,12 @@ export default class Diff extends Base {
         }
         Object.values(matchesBySdk).forEach((matches) => {
             matches.forEach((match) => {
+                const aliasedName = aliasMap[match.name]
+                if (match.isUnknown && aliasedName) {
+                    match.alias = match.name
```
ajwootto, line 142: do you think we should avoid mutation here? Technically these lists are being passed around in series so it shouldn't matter, but it could lead to something unexpected later maybe?

src/types/configFile.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
+}
+
+class CodeInsights {
+    @IsString({ each: true })
+    @IsOptional()
+    clientNames?: string[]
+
+    @IsOptional()
+    @IsObject()
```
ajwootto, line 42: is the validation here sufficient? Any way I could break things with some invalid configuration?

ajwootto: what if I provide a list instead, does that pass the IsObject validator?

laurawarr: an array will fail IsObject, I've added a custom validator to verify that the values in the map are strings

ajwootto (approved):

PR #38, opened 2022-02-03
add togglebot header svg

ajwootto:
image


laurawarr:
should this also include the usage of the icon?

ajwootto:
gonna do that in a followup PR since it felt weird to reference an image that doesn't exist yet

laurawarr (approved):

PR #43, opened 2022-02-15
Update togglebot icon for dark mode

laurawarr:
##  DevCycle Variable Changes:

🟢  0 Variables Added
🔴  0 Variable Removed


src/commands/diff/index.ts
```diff
@@ -253,8 +253,15 @@ export default class Diff extends Base {
         const subHeaderPrefix = this.useMarkdown ? '### ' : ''
 
         if (this.useMarkdown) {
-            this.log('\n## <img src="https://github.com/DevCycleHQ/cli/raw/main/assets/togglebot.svg" ' +
-                'height="31px" align="center"/> DevCycle Variable Changes:\n')
+            const lightTogglebot = 'togglebot.svg#gh-light-mode-only'
+            const darkTogglebot = 'togglebot-white.png#gh-dark-mode-only'
```
ajwootto, line 257: Should this one also be an SVG? Might end up a bit blurry since we're resizing it.

laurawarr: updated to an svg — the PR description uses the updated icon based on light/dark mode

ajwootto (approved):

PR #44, opened 2022-02-28
chore: Rename parse to parseDiffs

jsalaber:


src/utils/parsers/common.ts
```diff
@@ -306,7 +306,20 @@ export abstract class BaseParser {
         return results
     }
 
-    parseFile(_file: usage.File) {
-        return []
+    parseFile(_file: usage.File): FileMatch[] {
```
laurawarr, line 309: could remove the `_` from the `file` param, I'd just added that so it wouldn't complain about an unused param

src/utils/parsers/common.ts
```diff
@@ -305,7 +306,20 @@ export abstract class BaseParser {
...
+        const results: FileMatch[] = []
+
+        for (const line of _file.lines) {
+            const matches = this.getAllMatches(line.content)
```
laurawarr, line 313: are you planning to have a follow up PR to extend aggregateMultilineChunks to use here?

jsalaber: if we can use `aggregateMultilineChunks` to give us the chunk with the code, then yes

src/commands/usages/types.ts
```diff
@@ -0,0 +1,9 @@
+export type File = {
```
laurawarr, line 1: this file was replaced by `types.d.ts` on the `hackweek` branch so that it doesn't return that error when running the `usages` command

laurawarr (approved):

PR #48, opened 2022-03-01
feat: multi line method calls usage

jsalaber:
# Changes

- new function `extractUsageInformation` that takes the `File` object and the `MatchWithRange` to find the variable usage from the start and end indices.

### DIFF
--- a/src/utils/parsers/common.ts
+++ b/src/utils/parsers/common.ts
@@ -1,13 +1,8 @@
 import parse from 'parse-diff'
-import { VariableMatch, ParseOptions, VariableDiffMatch } from './types'
+import { VariableMatch, ParseOptions, VariableDiffMatch, Range, MultilineChunk, MultilineUsageChunk, VariableUsageMatch } from './types'
 import * as usage from '../../commands/usages/types'
 
-type Range = {
-    start: number
-    end: number
-}
-
-type MultilineChunk = {
+type MultilineDiffChunk = {
     content: string
     changes: ParsedChange[]
 }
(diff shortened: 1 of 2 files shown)


src/utils/parsers/common.ts
```diff
@@ -147,23 +142,41 @@ export abstract class BaseParser {
...
+        
+        for (const line of file.lines) {
+            index += line.content.length - 1
+            if (index >= match.range.start && index <= match.range.end) {
+                lines.push(line)
+            }
+        }
+
+        return {
+            content: lines.reduce((prev, curr) => prev + curr.content, ''),
```
laurawarr, line 157: were you thinking that this could be used to display the matched content? should we join with escaped newline characters so that we could display this formatted?

src/utils/parsers/common.ts
```diff
@@ -306,20 +319,25 @@ export abstract class BaseParser {
...
         for (const line of file.lines) {
-            const matches = this.getAllMatches(line.content)
-            for (const match of matches) {
-                results.push({
-                    fileName: file.name,
-                    line: match.range.start,
-                    name: match.name
-                })
-            }
+           fileContent = fileContent.concat(line.content)
```
laurawarr, line 326: in the diff implementation we skip any lines that start with a comment character, do you think something like that could work here too?

laurawarr (approved):

src/utils/parsers/common.ts
jsalaber: yes that was the idea, i can join them with new lines in another pr !

src/utils/parsers/common.ts
jsalaber: yeah we can do that too, but i had some work on a different branch handling single line comments

PR #49, opened 2022-03-01
feat: handle single line comments

jsalaber:
# Changes

- added check for start of line if it starts with a comment

# TODO:

- still needs to take in account comments added at the end of a line

### DIFF
--- a/src/utils/parsers/common.ts
+++ b/src/utils/parsers/common.ts
@@ -164,8 +164,8 @@ export abstract class BaseParser {
 
         return {
             content: lines.reduce((prev, curr) => prev + curr.content, ''),
-            start: lines[0].ln,
-            end: lines[lines.length - 1].ln
+            start: lines.length ? lines[0].ln : -1,
+            end: lines.length ? lines[lines.length - 1].ln : -1
         }
     }
 
@@ -339,12 +339,14 @@ export abstract class BaseParser {
 
         for (const match of matches) {
             const usage = this.extractUsageInformation(file, match)
-            result.push({
-                name: match.name,
-                line: usage.start,
-                fileName: file.name,
-                content: usage.content
-            })
+            if (usage.start !== -1 && usage.end !== -1) {
+                result.push({
+                    name: match.name,
+                    line: usage.start,
+                    fileName: file.name,
+                    content: usage.content
+                })
+            }
         }
 
         return result
--- a/src/utils/parsers/common.ts
+++ b/src/utils/parsers/common.ts
@@ -147,6 +147,7 @@ export abstract class BaseParser {
         let index = 0
     
         for (const line of file.lines) {
+<<<<<<< HEAD
             const trimmedContent = line.content.trim()
             if (!trimmedContent.length) continue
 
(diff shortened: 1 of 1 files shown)


src/utils/parsers/common.ts
```diff
@@ -148,15 +148,18 @@ export abstract class BaseParser {
         
         for (const line of file.lines) {
             index += line.content.length - 1
+            if (line.content.replace(/ /g, '').startsWith(this.commentCharacters[0])) {
```
laurawarr, line 151: You could use `.trim()` to handle any kind of whitespace character. Also, shouldn't this check for any of the commentCharacters, not just the first one?

src/utils/parsers/common.ts
```diff
@@ -148,15 +148,18 @@ export abstract class BaseParser {
...
                 lines.push(line)
             }
         }
 
         return {
             content: lines.reduce((prev, curr) => prev + curr.content, ''),
-            start: lines[0].ln,
-            end: lines[lines.length - 1].ln
+            start: lines.length ? lines[0].ln : -1,
+            end: lines.length ? lines[lines.length - 1].ln : -1
```
laurawarr, line 162: what's the case that lines will be empty?

src/utils/parsers/common.ts
```diff
@@ -148,15 +148,22 @@ export abstract class BaseParser {
         
         for (const line of file.lines) {
             index += line.content.length - 1
+            let isComment = false
+            for (const commentChar of this.commentCharacters) {
```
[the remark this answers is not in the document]
jsalaber: I think if we exclude it from the `fileContent` it will mess up the line numbers, e.g. excluding a single line comment will make the index of any variable used after be off by 1

src/utils/parsers/common.ts
```diff
@@ -163,7 +163,7 @@ export abstract class BaseParser {
         }
 
         return {
-            content: lines.reduce((prev, curr) => prev + curr.content, ''),
+            content: lines.reduce((prev, curr) => prev + (prev ? '\n' : '') + curr.content, ''),
```
laurawarr, line 166: could you not just use `lines.join('\n')` here?

jsalaber: the lines  array contains `LineItem` objects, not strings unfortunately

laurawarr: What if we get the content by using the start/end lines instead? Since this is using the filtered file it will be missing the comment lines. Instead we could say `content = file.lines.slice(start - 1, end).map(ln => ln.content).join('\n')`

laurawarr: That would also make it easy to add surrounding lines later

jsalaber: i changed it to just `lines.map(line => line.content).join('\n')`. It's simpler but also doesn't need to check the file for the lines since it already has the lines it needs to display

src/utils/parsers/common.ts
```diff
@@ -330,15 +350,16 @@ export abstract class BaseParser {
 
     parseFile(file: usage.File): VariableUsageMatch[] {
         const result: VariableUsageMatch[] = []
+        const filteredFile = this.getFilteredFile(file)
         let fileContent = ''
-        for (const line of file.lines) {
-            fileContent = fileContent.concat(line.content.trim()) 
+        for (const line of filteredFile.lines) {
+            fileContent = fileContent.concat(line.content)
```
laurawarr, line 356: oh! I think dropping the `trim` is causing the extra lines (since `extractUsageInformation` is still trimmin)

jsalaber: i'm a dummy

laurawarr (approved):

PR #51, opened 2022-03-02
feat: add buffer lines

jsalaber:
# Changes

- Added buffer variable that takes the context around the variable usage

### DIFF
--- a/src/utils/parsers/common.ts
+++ b/src/utils/parsers/common.ts
@@ -142,11 +142,11 @@ export abstract class BaseParser {
         )
     }
 
-    private extractUsageInformation(file: usage.File, match: MatchWithRange): MultilineUsageChunk {
+    private extractUsageInformation(file:usage.File, filteredFile: usage.File, match: MatchWithRange): MultilineUsageChunk {
         const lines: usage.LineItem[] = []
         let index = 0
     
-        for (const line of file.lines) {
+        for (const line of filteredFile.lines) {
             const trimmedContent = line.content.trim()
             if (!trimmedContent.length) continue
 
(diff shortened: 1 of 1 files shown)


src/utils/parsers/common.ts
```diff
@@ -162,8 +162,16 @@ export abstract class BaseParser {
...
+        const start = lines[0].ln
+        const end = lines[lines.length - 1].ln
+        const content = file.lines
+            .filter((line) => start - buffer <= line.ln && end + buffer >= line.ln)
+            .map(line => line.content)
+            .join('\n')
+
         return {
-            content: lines.map((line) => line.content).join('\n'),
+            content,
```
laurawarr, line 174: might as well use the start/end variables here

src/utils/parsers/common.ts
```diff
@@ -359,7 +367,7 @@ export abstract class BaseParser {
         const matches = this.getAllMatches(fileContent)
 
         for (const match of matches) {
-            const usage = this.extractUsageInformation(filteredFile, match)
+            const usage = this.extractUsageInformation(file, filteredFile, match)
```
laurawarr, line 370: what if we parsed the content here based on the output line numbers? rather than passing both the file and the filteredFile

laurawarr: or made it's own method to parse the content, given the file and the resulting line numbers

src/utils/parsers/common.ts
```diff
@@ -360,11 +361,18 @@ export abstract class BaseParser {
...
+                .filter((line) => usage.start - buffer <= line.ln && usage.end + buffer >= line.ln)
+                .map(line => line.content)
+                .join('\n')
+            
             result.push({
                 name: match.name,
                 line: usage.start,
                 fileName: file.name,
-                content: usage.content
+                content: usage.content,
```
laurawarr, line 374: do we need both types of content? I think just the buffered content may be sufficient

src/utils/parsers/common.ts
```diff
@@ -163,7 +163,7 @@ export abstract class BaseParser {
         }
 
         return {
-            content: lines.map((line) => line.content).join('\n'),
+            content: lines.map(line => line.content).join('\n'),
```
laurawarr, line 166: we could remove this content since we're not using it

PR #53, opened 2022-03-03
feat: update json output structure

jsalaber:
# Changes

- If user inputs json output, format result into JSON in the form of: 
```
[
    {
      "key": "string",
      "references": [
        {
          "codeSnippet": {
            "lineNumbers": {
              "start": 1,
              "end": 1
            },
            "content": "string"
          },
          "lineNumbers": {
            "start": 1,
            "end": 1
          },
          "fileName": "string",
          "language": "string"
        }
      ]
    }
  ]
```

### DIFF
--- a/src/utils/parsers/common.ts
+++ b/src/utils/parsers/common.ts
@@ -361,8 +361,8 @@ export abstract class BaseParser {
         for (const match of matches) {
             const range = this.extractUsageRange(filteredFile, match)
 
-            const bufferedStart = (range.start - buffer) < 0 ? 0 : range.start - buffer
-            const bufferedEnd = (range.end + buffer) > file.lines.length ? file.lines.length : range.end + buffer
+            const bufferedStart = Math.max(range.start - buffer, 0)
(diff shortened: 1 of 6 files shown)


src/utils/parsers/common.ts
```diff
@@ -361,6 +361,8 @@ export abstract class BaseParser {
         for (const match of matches) {
             const range = this.extractUsageRange(filteredFile, match)
 
+            const bufferedStart = (range.start - buffer) < 0 ? 0 : range.start - buffer
+            const bufferedEnd = (range.end + buffer) > file.lines.length ? file.lines.length : range.end + buffer
```
laurawarr, line 365:
```suggestion
            const bufferedStart = Math.max(range.start - buffer, 0)
            const bufferedEnd = Math.min(range.end + buffer, file.lines.length)
```
you could do this a little more simply with min/max

laurawarr (approved):

laurawarr:
oh, it looks like the test needs to be updated since we're no longer expecting the single-comment to be matched

PR #58, opened 2022-03-08
Add some management API functionality

andrewtraviss:
- Added SSO auth
- Added org switching
- Added project switching via command
- Added variable list/get/create commands
- Added feature list/get commands
- Switched help to use the multi-document format
- Put global flags into a separate flag category
- Updated the format of auth.yml to support multiple auth methods


laurawarr:
add `.DS_Store` to the gitignore?

src/api/features.ts
```diff
@@ -0,0 +1,46 @@
+import axios from 'axios'
+import { BASE_URL } from './common'
+export class Feature {
+    _id: string
```
laurawarr, line 4: another reason to get the api types into a module 👀

andrewtraviss: Yeah that would be really handy here.

andrewtraviss:
> add `.DS_Store` to the gitignore?

Done

laurawarr (approved):

PR #62, opened 2022-03-10
Hackweek - Code Usages

laurawarr:


jsalaber (approved):
maybe get adam to do a once over before merging

src/commands/usages/index.ts
```diff
@@ -0,0 +1,181 @@
...
+            description: 'Files to include when scanning for usages. By default all files are included. ' +
+                'Accepts multiple glob patterns.',
+            multiple: true
+        }),
+        'exclude': Flags.string({
+            description: 'Files to exclude when scanning for usages. By default all files are included. ' +
+                'Accepts multiple glob patterns.',
+            multiple: true
+        }),
+        'client-name': ClientNameFlag,
```
ajwootto, line 37: it seems like there's some potential to have a "code insights" base class here that the diff and usage commands can extend from, which itself extends from the base class?

src/commands/usages/index.ts
```diff
@@ -0,0 +1,181 @@
...
+                const aliasedName = aliasMap[match.name]
+                if (match.isUnknown && aliasedName) {
+                    match.alias = match.name
+                    match.name = aliasedName
+                    delete match.isUnknown
+                }
+                if (!match.isUnknown) {
+                    matchesByVariable[match.name] ??= []
+                    matchesByVariable[match.name].push(match)
+                    matchesByVariable[match.name] = uniqBy(
```
ajwootto, line 131: minor thing but it seems inefficient to be redoing the uniqBy on each iteration. Maybe you could use a set here? Or do the uniqBy at the end for each array in the object?

test/utils/parsers/android.test.ts
```diff
@@ -1,6 +1,6 @@
-import { executeFileDiff } from '../../../../src/utils/diff/fileDiff'
```
ajwootto, line 1: wouldnt these tests still belong in the "diff" directory since they're specifically testing parsing of diffs?

src/utils/parsers/types.ts
```diff
@@ -0,0 +1,34 @@
+export type VariableMatch = {
```
ajwootto, line 1: some comments on these would be helpful

src/utils/parsers/BaseParser.ts
```diff
@@ -31,31 +22,47 @@ export type MatchResult = {
...
-    }
-
-    parseContent() {
-        return this.type === 'normal'
-            ? this.content.trim()
-            : this.content.slice(1).trim()
+    constructor(line: usage.LineItem | parse.Change) {
+        this.content = line.content
+        this.parsedContent = this.content.trim()
+        this.ln = (line as usage.LineItem).ln
```
ajwootto, line 38: whats going on here?

src/utils/parsers/BaseParser.ts
```diff
@@ -246,61 +209,121 @@ export abstract class BaseParser {
...
     }
 
     /**
-     * Given an array of matches, find the first change object corresponding to the match.
-     * Also verify that the match is associated with at least one add/delete change object.
+     * Get all lines within the range of the match
      */
-    private formatMatches(file: parse.File, matches: MatchWithRange[], { changes }: MultilineChunk) {
-        const results: VariableMatch[] = []
+    private getMatchingLines<T>(match: MatchWithRange, lines: ParsedLine[]): T[] {
```
ajwootto, line 233: how come this needs a generic type arg if it's just type casting to the provided type at the end? Doesn't that mean there's really only one type here?

test/utils/parsers/android.test.ts
laurawarr: I'd intended to add other tests to these, but didn't get around to it — I'll move them back for now

src/utils/parsers/BaseParser.ts
laurawarr: this handles the result being either an array of `ParsedLine` or an array of `ParsedChangeLine` in the diff case

ajwootto: but the actual data being returned doesn't look any different so why allow the type of its return to be overridden like this?

src/utils/parsers/BaseParser.ts
laurawarr: The `ParsedLine` class is used as a common structure to work with both diff/file lines. The only problem is when extending this in `ParsedChangeLine`, `super()` is unhappy because normal diff lines don't have a `ln` property

src/utils/parsers/BaseParser.ts
laurawarr: the returned object type matches what's passed in, `ParsedChangeLine` has additional properties that are used later

ajwootto: but what's passed in is always of type "ParsedLine", and so the return is always an array of "ParsedLine"

ajwootto: if you're trying to represent two different call signatures here then you should probably use function overloading 
https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads

laurawarr: that's cool, but it doesn't seem to work here, even with:
```
    private getMatchingLines(match: MatchWithRange, lines: ParsedChangeLine[]): ParsedChangeLine[];
    private getMatchingLines(match: MatchWithRange, lines: ParsedLine[]): ParsedLine[] {
```
typescript complains:
`Argument of type 'ParsedLine[]' is not assignable to parameter of type 'ParsedChangeLine[]'`

laurawarr: ParsedChangeLine extends ParsedLine, so typescript still considers that a ParsedLine

laurawarr: scratch that previous comment — just realized the function signature can't be called and it needs another overload

PR #63, opened 2022-03-14
[DVC-2660] Change the default port

andrewtraviss:
Port 5000 is used by upnp, so the default needs to be changed to an
unassigned port. There is support for overriding the port, but the
oauth flow requires us to specify callback urls against specific port
numbers, so it's necessary to have a fixed list of supported ports.

### DIFF
--- a/src/api/ssoAuth.ts
+++ b/src/api/ssoAuth.ts
@@ -14,6 +14,14 @@ type OauthResponse = {
         access_token: string
     }
 }
+const SUPPORTED_PORTS = [
+    80,
+    8080,
+    2194,
+    2195, 
+    2196
+]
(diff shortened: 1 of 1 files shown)


src/api/ssoAuth.ts
```diff
@@ -14,6 +14,14 @@ type OauthResponse = {
...
     }
 }
+const SUPPORTED_PORTS = [
+    80,
+    8080,
+    2194,
+    2195, 
+    2196
+]
+const PORT=Number.parseInt(process.env.PORT || '2194')
```
jonathannorris, line 24:
```suggestion
const PORT = Number.parseInt(process.env.PORT || '2194')
```

jonathannorris (approved):

PR #65, opened 2022-03-14
Map file extensions to language

laurawarr:
Using the parser identity to determine the language does not work in the CustomParser case (returns 'custom' as the language). Instead, map the file extension to the language

### DIFF
--- a/src/utils/parsers/index.ts
+++ b/src/utils/parsers/index.ts
@@ -36,3 +36,18 @@ export const PARSERS: Record<string, (typeof NodeParser)[]> = {
     swift: [IosParser],
     php: [PhpParser]
 }
+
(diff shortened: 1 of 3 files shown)


src/utils/parsers/index.ts
```diff
@@ -36,3 +36,18 @@ export const PARSERS: Record<string, (typeof NodeParser)[]> = {
     swift: [IosParser],
     php: [PhpParser]
 }
+
+export const LANGAUGE_MAP: Record<string, string> = {
```
jsalaber, line 40: language tyyypo

jsalaber (changes requested):

jsalaber (approved):

PR #70, opened 2022-03-24
Dvc 2797 config files shouldn't be stored in current directory by default

andrewtraviss:
I also added logout so that I could easily test logging in and out again

### DIFF
--- a/src/commands/logout/index.ts
+++ b/src/commands/logout/index.ts
@@ -1,0 +1,16 @@
+import fs from 'fs'
+import { togglebotSleep } from '../../ui/togglebot'
+import Base from '../base'
+
+export default class Logout extends Base {
+    static hidden = false
+    static description = 'Discards any auth configuration that has been stored in the auth configuration file.'
+    static examples = []
+
+    public async run(): Promise<void> {
+        if (fs.existsSync(this.authPath)) {
+            fs.rmSync(this.authPath)
+        }
+        console.log(togglebotSleep) 
+    }
+}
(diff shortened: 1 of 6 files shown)


src/commands/logout/index.ts
```diff
@@ -0,0 +1,16 @@
+import fs from 'fs'
+import { togglebotSleep } from '../../ui/togglebot'
+import Base from '../base'
+
+export default class Logout extends Base {
+    static hidden = false
+    static description = 'Log in through the DevCycle Universal Login. This will open a browser window'
```
ajwootto, line 7: think this needs an update

src/ui/togglebot.ts
```diff
@@ -30,4 +30,19 @@ export const togglebotWink = chalk.blue(`
   "-.,_______________,.-"
                              
          \`-.___.-\`
+`)
+
+export const togglebotSleep = chalk.blue(`
```
ajwootto, line 35: lol

src/commands/logout/index.ts
```diff
@@ -0,0 +1,16 @@
...
+    static examples = []
+
+    public async run(): Promise<void> {
+        if (fs.existsSync(this.authPath)) {
+            fs.rmSync(this.authPath)
+        }
+        console.log(togglebotSleep)
```
ajwootto, line 14: do we not want all logs to go through the built-in logging system? I imagine part of the reason for doing that is that it would let us properly implement a "--silent" flag or change the amount of logging output

andrewtraviss: When we do that we should use our own logging function. The oclif solution requires any code that wants to write to the log to have a reference to the Command, which is probably my biggest gripe about oclif.

PR #74, opened 2022-06-08
DVC-3490 Separate repo config from user config

andrewtraviss:
User configuration should not be stored in the repo, so we need
separate configuration files to handle repo-configuration that the
diff and usages commands rely on from cli user configuration and
session data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,8 @@ The CLI can be customized in several ways using command-line args or by creating
 * [Authentication](#authentication)
 * [Usage](#usage)
 * [Commands](#commands)
-* [Configuration](#configuration)
+* [Command Topics](#command-topics)
+* [Repo Configuration](#repo-configuration)
 <!-- tocstop -->
 # Setup
 ## Install the CLI:
@@ -32,8 +33,7 @@ This process will open browser windows to interact with the DevCycle universal l
 To switch organizations once logged in, the [org command](docs/org.md) can be used.
 ## Using Client Credentials
 ### Credentials File
-Create a subdirectory inside the directory where you're running the CLI called `.devcycle`, then inside that directory
-create an `auth.yml` file with the following contents:
+Check the [oclif docs](https://oclif.io/docs/config) to find where the default `configDir` is located. Within that directory, create a `devcycle` directory, if it does not already exist. Inside of the `devcycle` directory, create an `auth.yml` file with the following contents:
 
 ```yaml
 clientCredentials:
@@ -41,13 +41,13 @@ clientCredentials:
   client_secret: <your client secret>
 ```
 This file should **not** be checked in to version control.
+## Project Selection
 
-You also need to specify the default project ID for the CLI to use. 
+You also need to specify the default project ID for the CLI to use.
 
-This can be set using the [project select command](docs/project.md#dvc-projects-select) or by manually updating the [configuration](#configuration) file:
-```yaml
-project: <your project id>
-```
+If there is a repo configuration file, the [`dvc diff`](docs/diff.md) and [`dvc usages`](docs/usages.md) commands will use the project defined there.
+
+Otherwise, this is chosen during login or set using the [project select command](docs/project.md#dvc-projects-select)
 
 ### Environment Variables
 Set the following environment variables:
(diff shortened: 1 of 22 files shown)


src/commands/base.ts
```diff
@@ -17,13 +17,17 @@ export default abstract class Base extends Command {
...
         'config-path': Flags.string({
-            description: 'Override the default location to look for a config.yml file',
+            description: 'Override the default location to look for the user.yml file',
             helpGroup: 'Global'
         }),
         'auth-path': Flags.string({
             description: 'Override the default location to look for an auth.yml file',
             helpGroup: 'Global'
         }),
+        'repo-path': Flags.string({
```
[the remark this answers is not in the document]
andrewtraviss: I was torn between the two since this is not consistent with the other path flags. I'll go with your suggestion.

src/commands/base.ts
```diff
@@ -114,10 +133,15 @@ export default abstract class Base extends Command {
             this.configPath = flags['config-path']
         }
 
-        this.configFromFile = this.loadConfig(this.configPath)
+        if (flags['repo-config-path']) {
+            this.repoPath = flags['repo-config-path']
```
laurawarr, line 137: it's probably worth updating the property name to match, just for clarity
```suggestion
            this.repoConfigPath = flags['repo-config-path']
```

README.md
```diff
@@ -32,8 +33,7 @@ This process will open browser windows to interact with the DevCycle universal l
 To switch organizations once logged in, the [org command](docs/org.md) can be used.
 ## Using Client Credentials
 ### Credentials File
-Create a subdirectory inside the directory where you're running the CLI called `.devcycle`, then inside that directory
-create an `auth.yml` file with the following contents:
+Create a `.devcycle` subdirectory inside your home directory. Inside that directory create an `auth.yml` file with the following contents:
```
laurawarr, line 36: is this correct? I think by default it uses `.config/oex`, we should also document the repo config

andrewtraviss: You're right, but I believe this is platform-specific. I should link out to the oclif readme section that covers default config paths.

I'll also add an output line whenever the config is saved to say where it was saved.

andrewtraviss: Ah, discovered a configuration issue. "oex" should be an application-specific folder name. Changing that to `devcycle`

README.md
```diff
@@ -32,8 +33,7 @@ This process will open browser windows to interact with the DevCycle universal l
 To switch organizations once logged in, the [org command](docs/org.md) can be used.
 ## Using Client Credentials
 ### Credentials File
-Create a subdirectory inside the directory where you're running the CLI called `.devcycle`, then inside that directory
-create an `auth.yml` file with the following contents:
+Check the [oclif docs](https://oclif.io/docs/config) to find where the default `configDir` is located. Within that directory, create a `devcycle` directory, if it does not already exist. Inside of the `devcycle` directory, create an `auth.yml` file with the following contents:
```
ajwootto, line 36: not sure if this is addressed by Laura's other comment, but it doesn't seem like we should need to point people to oclif to know where the configuration is stored, we should control it ourselves

README.md
```diff
@@ -77,344 +76,18 @@ USAGE
...
-  --config-path=<value>    Override the default location to look for a config.yml file
-  --no-api                 Disable API-based enhancements for commands where authorization is optional. Suppresses
-                           warnings about missing credentials.
-  --project=<value>        Project key to use for the DevCycle API requests
-
-DESCRIPTION
-  Update a Variable.
-```
 <!-- commandsstop -->
 # Configuration
```
ajwootto, line 92: this section needs to be updated doesn't it?

src/commands/base.ts
```diff
@@ -46,14 +51,16 @@ export default abstract class Base extends Command {
     token = ''
     projectKey = ''
     authPath = path.join(this.config.configDir, 'auth.yml')
-    configPath = path.join(this.config.configDir, 'config.yml')
+    configPath = path.join(this.config.configDir, 'user.yml')
+    repoConfigPath = '.devcycle/config.yml'
```
ajwootto, line 55: shouldnt this be read from the repo-config-path flag? And have a default value set there?

andrewtraviss: These are the default values. If the flags exist, they override these values later. (Lines 130-140)

README.md
andrewtraviss: We're using oclif's built-in "configDir" as they recommend, which puts the configuration in the right place for user-specific configuration depending on what platform you're running the CLI on.

README.md
andrewtraviss: Yep. Most of it still seems to be accurate, it just only applies to the diff and usages commands.

src/commands/logout/index.ts
```diff
@@ -1,4 +1,5 @@
 import fs from 'fs'
+import { successMessage } from '../../ui/output'
```
laurawarr, line 2: could this be removed?

laurawarr (approved):
look good 👍 just one import that looks like it could be removed

PR #78, opened 2022-07-25
Hack week variable cleanup

laurawarr:


andrewtraviss (commented):
if the variable key isn't provided, and the user is authenticated, you could use a `variablePrompt` to get the variable key 
```
        const { variable } = await inquirer.prompt([variablePrompt], {
            token: this.token,
            projectKey: this.projectKey
        })
```
That snippet will get the list of all variables in the project, present the user with a prompt to choose it from a list, and return the Variable object for the chosen variable.

andrewtraviss (approved):

PR #79, opened 2022-07-27
Hack week updates to better support tooling integration

andrewtraviss:
- Implemented `dvc repo init` to initialize the current repo with a repo configuration file, using the sso login process to retrieve valid organization and project ids
- Implemented `dvc alias add` to add a new alias to the repo configuration
- Added `dvc login again` to log in using only stored login selections, and skip the first personal login step
- Added a `--headless` flag for all commands that eliminates most of the output from the CLI to make it easier for other tools to rely on its output.
- In headless mode, `dvc status` outputs status information in a json format
- Added support to `dvc login sso` and `dvc org` to accept the organization name from a flag or from the repo configuration
- Moved most of the SSO login code into a shared base AuthCommand
- Fixed a bug where the SSO login would break if the two stage login happened too quickly
- Fixed a bug where `dvc status` was looking at the wrong path for the user configuration


laurawarr (approved):
There's a handful of linter errors to fix, but otherwise looks good 👍

laurawarr (approved):

PR #83, opened 2022-08-22
Add support for react useVariableValue method

laurawarr:
Update ReactParser to handle `useVariableValue` method, this will be used for the `dvc usages` and `dvc diff` commands


src/utils/parsers/react/index.ts
```diff
@@ -3,10 +3,18 @@ import { BaseParser } from '../BaseParser'
 const variableNameCapturePattern = /([^,)]*)/
 const defaultValueCapturePattern = /(?:[^),]*|{[^}]*})/
 
+const supportedMethods = [
+    'useDVCVariable',
```
ajwootto, line 7: this one is already deprecated..still worth adding?

ajwootto (approved):

src/utils/parsers/react/index.ts
laurawarr: If we're sure no one is using an old version of the sdk — Mark was testing these commands and his codebase was still using `useDVCVariable`

laurawarr: So I would say it doesn't hurt to have it for now. It's very easy to remove it later

PR #85, opened 2022-08-30
add environments to the cli

andrewtraviss:
Adding support for environments to the CLI is a necessary prerequisite of working with Feature Configurations.

This also adds the ability to retrieve SDK keys via the CLI.


docs/envs.md
```diff
@@ -0,0 +1,93 @@
+`dvc envs`
```
laurawarr, line 1: is there a difference between `dvc envs` and `dvc environments`?

src/api/environments.ts
```diff
@@ -0,0 +1,127 @@
+import axios from 'axios'
+import { BASE_URL } from './common'
+import { IsIn, IsNotEmpty, IsOptional, IsString } from 'class-validator'
+import { string } from '@oclif/core/lib/flags'
```
laurawarr, line 4: based on the github comment, could this be removed?

docs/envs.md
andrewtraviss: Hmm envs shouldn't be here anymore.

laurawarr (approved):

PR #86, opened 2022-09-28
Fix show-regex for custom match patterns

laurawarr:
- Add `printRegexPattern` method specific to the custom parser
- Update CustomParser `identity` property to include file extension so that each CustomParser has a unique identity. This was causing only one match pattern to be logged


src/utils/parsers/custom.ts
```diff
@@ -13,9 +13,15 @@ export class CustomParser extends BaseParser {
             throw new Error(`No match pattern for ${extension}`)
         }
 
+        this.identity = `custom ${extension}`
```
ajwootto, line 16: aren't you overwriting the identity field being set to "custom" above? Is that going to cause problems elsewhere? We use the identity to figure out which parser to use don't we

laurawarr: Yeah, the identity "custom" isn't unique, but we create a custom parser for each extension. This was an issue when the `show-regex` flag is enabled and we're printing values to the console, we build a map to track which parsers have been printed, and since all custom parsers have the same identity only one of them would be logged

ajwootto: maybe just remove the line that's assigning it to `custom` then since that value should never be used anyway?

ajwootto (approved):

PR #94, opened 2022-11-23
Mock file system in tests

andrewtraviss:
I need to add some more tests to this, but what is done is ready for review.


src/commands/diff/index.ts
```diff
@@ -79,6 +79,7 @@ export default class Diff extends Base {
     useHTML = false
 
     public async run(): Promise<void> {
+        console.error({cwd: process.cwd()})
```
laurawarr, line 82: should this still be here?

andrewtraviss: Nope

src/commands/login/sso.ts
```diff
@@ -11,8 +10,11 @@ export default class LoginSSO extends AuthCommand {
 
     public async run(): Promise<void> {
         const ssoAuth = new SSOAuth(this.writer)
-        this.token = await ssoAuth.getAccessToken()
-        storeAccessToken(this.token, this.authPath)
+        const accessToken = await ssoAuth.getAccessToken()
+        this.token = accessToken
```
suthar26, line 14: is this being used anywhere, I see that `authConfig.sso.accessToken` is being used instead

src/commands/repo/init.ts
```diff
@@ -10,15 +9,18 @@ export default class InitRepo extends AuthCommand {
...
         }
 
-        this.repoConfig = await this.updateRepoConfig({})
+        this.dvcConfig.updateRepoConfig({})
 
         const ssoAuth = new SSOAuth(this.writer)
-        this.token = await ssoAuth.getAccessToken()
-        storeAccessToken(this.token, this.authPath)
+        const accessToken = await ssoAuth.getAccessToken()
+        this.token = accessToken
```
suthar26, line 20: same here unless if I am missing anything

suthar26 (changes requested):

src/commands/login/sso.ts
andrewtraviss: All of the other commands set and use this.token, and this ensures that bugs will never be introduced in the case that someone assumes that this is true for all commands.

laurawarr (approved):

suthar26 (approved):

PR #99, opened 2023-01-17
[feat] remove node version limitation to test with GH Actions

suthar26:


suthar26:
manual testing with local build using node version 19.4.0 across all commands result in the same output as the production version. (some things are broken in prod version, which will be addressed in other tickets)

laurawarr:
we should do the same tests with node 18, I believe that was the version that issues were reported

suthar26:
will test after andrew's changes are merged

PR #105, opened 2023-03-27
[chore] Add dvc-referrer headers to API requests

laurawarr:


src/api/common.ts
```diff
@@ -1,2 +1,17 @@
+import axios from 'axios'
+
 export const BASE_URL = 'https://api.devcycle.com/'
 export const AUTH_URL = 'https://auth.devcycle.com/'
+
+export const setDVCReferrer = (
+    command: string = 'unknown',
+    version: string,
+    caller: string = 'cli'
```
chantalchung, line 9: should caller only be set if the cli was called from an integration? or do we want to default to `cli`?

laurawarr: I think if we default to caller being the same as the source then we can always use "caller" to understand which tools developers are interacting with — that being said, we could handle the defaulting on the API side

chantalchung: I think it's fine to handle it here

chantalchung (approved):

PR #106, opened 2023-03-28
[chore] accept optional caller param

chantalchung:
- add an optional `caller` param that is an enum of: `github | bitbucket | cli` to determine which integration is calling the CLI


src/commands/base.ts
```diff
@@ -51,13 +51,19 @@ export default abstract class Base extends Command {
...
+            options: ['github', 'bitbucket', 'cli']
+        }),
     }
 
     token = ''
     projectKey = ''
     authPath = path.join(this.config.configDir, 'auth.yml')
     configPath = path.join(this.config.configDir, 'user.yml')
     repoConfigPath = '.devcycle/config.yml'
+    caller = ''
```
laurawarr, line 66: This should either be `undefined` or default to `cli`. `setDVCReferrer` won't default to `cli` if caller is passed as an empty string (the default param is only used if the passed param isn't set)

chantalchung: updated to default to `cli`

laurawarr (approved):

PR #108, opened 2023-03-28
[chore] update caller param options

chantalchung:
- update caller to accept: `github.pr_insights | github.code_usages | bitbucket.pr_insights | bitbucket.code_usages | cli`


docs/alias.md
```diff
@@ -12,24 +12,42 @@ Add a variable alias to the repo configuration
 ```
 USAGE
   $ dvc alias add [--config-path <value>] [--auth-path <value>] [--repo-config-path <value>] [--client-id
-    <value>] [--client-secret <value>] [--project <value>] [--no-api] [--headless] [--caller github|bitbucket|cli]
-    [--alias <value>] [--variable <value>]
+    <value>] [--client-secret <value>] [--project <value>] [--no-api] [--headless] [--caller
```
suthar26, line 15: could we take this out to a central page so we it can be managed from one place?

chantalchung: not sure how, these docs are auto-generated

suthar26 (approved):

PR #110, opened 2023-04-06
feat: create project

nikolasleblanc:
Adds a `projects create` to the CLI, prompting the user for a unique project id, a project name, and a description.


src/commands/projects/create.ts
```diff
@@ -0,0 +1,16 @@
+import { createProject, CreateProjectParams } from '../../api/projects'
+import { descriptionPrompt, keyPrompt, namePrompt } from '../../ui/prompts'
+import CreateCommand from '../createCommand'
+
+export default class CreateVariable extends CreateCommand<CreateProjectParams> {
```
laurawarr, line 5:
```suggestion
export default class CreateProject extends CreateCommand<CreateProjectParams> {
```

laurawarr (approved):
One typo, but otherwise looks good 👍

PR #130, opened 2023-05-25
[chore] add apiClient to wrap all api calls

suthar26:
updated oclif to `v3^`
- use apiClient everywhere to make error handling easier
some are formatting changes thanks to my auto formatter


README.md
laurawarr, line 1: this file is auto generated by oclif, are these changes from your local linter?

src/api/apiClient.ts
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+/* eslint-disable no-console */
+import axios, { AxiosError } from 'axios'
+
+export const BASE_URL = 'https://api.devcycle.com/'
+// export const BASE_URL = 'http://localhost:4001/'
```
laurawarr, line 6: leftover comment from testing?

src/api/apiClient.ts
```diff
@@ -0,0 +1,40 @@
...
+export const AUTH_URL = 'https://auth.devcycle.com/'
+
+export const apiClient = axios.create({
+    baseURL: `${BASE_URL}/v1/api`,
```
laurawarr, line 10: is this correct? our api doesn't have `/api` in the path

src/api/apiClient.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+apiClient.interceptors.response.use(
+    (response) => {
+        return response
+    },
+    (error: AxiosError) => {
+        if (error.response?.status === 401) {
+            window.dispatchEvent(new Event('sessionExpired'))
+        }
```
laurawarr, line 33: what is this event used for?

src/api/common.ts
```diff
@@ -3,7 +3,7 @@ import axios from 'axios'
 export const BASE_URL = 'https://api.devcycle.com/'
 export const AUTH_URL = 'https://auth.devcycle.com/'
 
-export const setDVCReferrer = (
+export const setDVCReferrerOLD = (
```
laurawarr, line 6: any reason to leave this here?

src/api/environments.ts
```diff
@@ -57,14 +57,12 @@ export const createEnvironment = async (
...
     const url = new URL(`/v1/projects/${project_id}/environments`, BASE_URL)
-    const response = await axios.post(url.href,
-        params,
-        {
-            headers: {
-                'Content-Type': 'application/json',
-                Authorization: token,
-            },
-        })
+    const response = await apiClient.post(url.href, params, {
```
laurawarr, line 60: with a common client we shouldn't need to pass the base url here anymore

package.json
```diff
@@ -46,6 +46,7 @@
     "@types/mocha": "^9.0.0",
     "@types/node": "^16.9.4",
     "@types/node-emoji": "^1.8.1",
+    "@types/tmp": "^0.2.3",
```
laurawarr, line 49: is this packaged used in this PR?

README.md
laurawarr: I'm wrong about this one — the auto generated files are in docs now 👀

src/api/apiClient.ts
suthar26: yeah this was not properly tested since we were passing BASE_URL and is fixed now

src/api/apiClient.ts
suthar26: not needed, copy paste error, should be error messaged shown to user

src/api/common.ts
suthar26: forgot to remove

package.json
suthar26: don't need, i think because of node version mismatch it was complaining but when using it with `nvm use` it is not required

.github/workflows/tests.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-node@v1
       with:
-        node-version: '14'
+        node-version: '18.16.0'
```
laurawarr, line 14: we should also be able to bump both of these actions to v3

src/api/apiClient.ts
```diff
@@ -0,0 +1,35 @@
+import axios, { AxiosError } from 'axios'
+import { BASE_URL } from './common'
+
+export const apiClient = axios.create({
+    baseURL: `${BASE_URL}`,
```
laurawarr, line 5: tiny thing, but this doesn't need to be a string template
```suggestion
    baseURL: BASE_URL,
```

src/api/environments.ts
```diff
@@ -56,15 +55,12 @@ export const createEnvironment = async (
...
-            headers: {
-                'Content-Type': 'application/json',
-                Authorization: token,
-            },
-        })
+    const response = await apiClient.post(`/v1/projects/${project_id}/environments`, params, {
+        headers: {
+            'Content-Type': 'application/json',
+            Authorization: token,
+        },
```
laurawarr, line 62: doesn't necessarily need to be handled in this PR, but it'd be great to move this to the api client so that we don't need to add these headers to every request

laurawarr (approved):

PR #137, opened 2023-06-05
[DVC-7562] [DVC-7618] [DVC-7619] [fix] add proper headless functionality to create environment

suthar26:


src/commands/environments/create.ts
```diff
@@ -1,30 +1,98 @@
...
+            description: 'Description for display in the dashboard.',
+        }),
+        'type': Flags.string({
+            description: 'The type of environment',
+            options: environmentTypes
+        }),
+    }
 
     public async run(): Promise<void> {
+        const { flags } = await this.parse(CreateEnvironment)
```
laurawarr, line 37: could we not move this to the CreateCommand and skip the prompt if the flag value is set or if the headless flag is set?

laurawarr: the command should handle returning any errors if the required fields are missing, but the common create prompt should be able to handle parsing and returning the parameter values

oclif.manifest.json
```diff
@@ -1 +1,2950 @@
...
\ No newline at end of file
+{
```
chantalchung, line 1: it looks like your oclif manifest is making the `args` objects instead of arrays too (so args are not working properly)
ex:
image

src/ui/prompts/featurePrompts.ts
```diff
@@ -18,7 +18,7 @@ export const featureChoices = async (input: Record<string, any>):Promise<Feature
 }
 
 export const featurePrompt = {
-    name: '_feature',
```
chantalchung, line 21: is there a reason for renaming this? might have to fix it in `variations/get` and `variations/list` since the featurePrompt is being used there too

oclif.manifest.json
suthar26: I did not change the cleanup command, and I believe kaushal's ticket should take care of the args not working as expected?

src/commands/variables/create.ts
```diff
@@ -14,7 +16,18 @@ import CreateCommand from '../createCommand'
...
-    
+
+    static flags = {
+        ...CreateCommand.flags,
+        'type': Flags.string({
+            description: 'The type of variable',
+            options: variableTypes
+        }),
+        'feature': Flags.string({
+            description: 'The key of the feature to create the variable for'
```
chantalchung, line 27: this can be feature key or id right?

suthar26: yes, fixed

src/ui/prompts/featurePrompts.ts
suthar26: ahh, will fix that. it is because I am matching the flag name to the prompt key

docs/projects.md
```diff
@@ -14,7 +14,13 @@ Create a new Project
...
 USAGE
   $ dvc projects create [--config-path <value>] [--auth-path <value>] [--repo-config-path <value>] [--client-id
-    <value>] [--client-secret <value>] [--project <value>] [--no-api] [--headless]
+    <value>] [--client-secret <value>] [--project <value>] [--no-api] [--headless] [--key <value>] [--name <value>]
+    [--description <value>]
+
+FLAGS
+  --description=<value>  Description for display in the dashboard.
+  --key=<value>          Unique ID to refer to the environment.
+  --name=<value>         Human readable name of the environment.
```
chantalchung, line 23: could these be confusing flag names if they're just `key` and `name` but refer to the environment key and names instead of project key and names? it makes sense for the environments commands but not here? I think it was added in the `createCommand.ts` file but could it just be in environments instead?

suthar26: yeah need to change the description of each flag, but they refer to the actual object's name and key

chantalchung (approved):

PR #138, opened 2023-06-05
[fix] select org from flag before prompting projects select

chantalchung:
- select organization when the flag is passed into the `projects select` command
cli-projects-select-org


suthar26:
tested locally by running the `projects select` command with --org flag

suthar26 (approved):

laurawarr (approved):

PR #142, opened 2023-06-06
Add variations get command

emmawillis:
https://github.com/DevCycleHQ/cli/assets/25067948/7e8e8d02-1770-4fd7-aa5c-9e182bd8fce6

(`get.ts` is exactly the same as `list.ts` but it directly returns the whole variables list instead of mapping it to only return the keys)

### DIFF
--- a/docs/variations.md
+++ b/docs/variations.md
@@ -3,7 +3,30 @@
 
 
 
+* [`dvc variations get [FEATURE]`](#dvc-variations-get-feature)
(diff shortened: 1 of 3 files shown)


chantalchung (approved):

chantalchung (changes requested):
I just tested locally but had to run `yarn build` (which reformatted the oclif manifest json file) otherwise I was getting the command not found error, it looks like the oclif manifest should be formatted in 1 line? 🤔 
image

chantalchung (approved):

PR #146, opened 2023-06-08
Replace axios with zodios

emmawillis:
- replaced the axios apiClient with zodios apiClient 
    - we still need to use axios directly for `fetchVariables` to get the count header, and also for the organizations endpoint because organizations are not in the swagger docs for some reason 
- added `generate-zodios-client` script which pulls the swagger docs and uses them to set up `zodClient.ts`
    - note: i'm not sure how/when we'll want to run this, but we'll need to run it regularly to get updates to the backend 
- removed explicit return types on all the api functions because they're inferred from the zodios client now
- exported the types directly from the zodClient and used them anywhere we used to use our own types 
- fixed up types wherever there were mismatches between the zod schema and our code

Note: i had to add  `validate: 'request'` as an option for now to prevent the zodios client trying to validate the response, because almost all of the endpoints have some small inconsistency between the actual return type and the swagger docs, so zodios will throw when the response doesn't pass validation

testing: i tested the commands to get and update all the objects


src/api/organizations.ts
```diff
@@ -7,8 +7,7 @@ export type Organization = {
 }
 
 export const fetchOrganizations = async (token: string): Promise<Organization[]> => {
-    const url = '/v1/organizations'
-    const response = await apiClient.get(url, {
+    const response = await axiosClient.get('/v1/organizations', { // TODO why does swagger not have /v1/organizations?
```
chantalchung, line 10: I think it's because the `organizations` API is part of the internal docs, is there something we can do there?

scripts/generate-zodios-client.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+url="https://api.devcycle.com/swagger.json"
+temp_file=$(mktemp)
+curl -s "$url" -o "$temp_file"
+openapi-zod-client "$temp_file" --with-alias --export-schemas --prettier "prettier.config.js" -o "src/api/zodClient.ts"
+rm "$temp_file"
```
chantalchung, line 6: maybe the script should be run before we do releases?

src/api/environments.ts
```diff
@@ -14,18 +14,6 @@ export const sdkTypes = [
...
-    name?: string
-    key: string
-    type: 'development' | 'staging' | 'production' | 'disaster_recovery'
-    sdkKeys: Record<string, APIKey[]>
-    description?: string
-    createdAt: Date
-    updatedAt: Date
-}
-
 export class CreateEnvironmentParams {
```
chantalchung, line 17: should we be replacing these `CreateEnvironmentParams`, `CreateProjectParams`, etc. with the types in the zodClient instead? like `CreateEnvironmentDto` and `CreateProjectDto`?

emmawillis: made a ticket to handle this! https://taplytics.atlassian.net/browse/DVC-7720

src/api/organizations.ts
emmawillis: yeah i'm thinking they should just be in the pubic docs anyway since they're public endpoints? i created a ticket to figure it out here https://taplytics.atlassian.net/browse/DVC-7723

scripts/generate-zodios-client.sh
emmawillis: i'll create a separate ticket for this one too, there are some issues with the openapi-zod-client using outdated types so after you run it you have to manually fix a couple of things :/ so its not ready to be fully automated yet

PR #161, opened 2023-06-14
DVC-7749 Update `targeting get` to prompt for environment

laurawarr:
- Update `dvc targeting get` to include prompt for environment:
If a feature is specified, do not use any prompts. Otherwise prompt for feature and environment
 
- Add tests for `dvc targeting get` command
- Small fixes to auth error message and variation list test

dvc-targeting-get

### DIFF
--- a/src/commands/targeting/get.test.ts
+++ b/src/commands/targeting/get.test.ts
@@ -1,0 +1,127 @@
+import { expect } from '@oclif/test'
+import inquirer from 'inquirer'
+import { dvcTest } from '../../../test-utils'
+import { BASE_URL } from '../../api/common'
+
+describe('targeting get', () => {
+    const projectKey = 'test-project'
+    const featureKey = 'test-feature'
+    const authFlags = [
+        '--project', projectKey,
+        '--client-id', 'test-client-id',
+        '--client-secret', 'test-client-secret'
+    ]
+
+    const mockTargeting = [
+        {
+            '_feature': '642d9de38bbfc09ad0657e05',
+            '_environment': '637cfe8195279288bc08cb61',
+            '_createdBy': 'google-oauth2|112946554154333301873',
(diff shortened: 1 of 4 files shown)


src/commands/targeting/get.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    dvcTest()
+        .nock(BASE_URL, (api) => api
+            .get(`/v1/projects/${projectKey}/features/${featureKey}/configurations`)
+            .reply(200, mockTargeting)
+        )
+        .stdout()
+        .command(['targeting get', featureKey, ...authFlags])
+        .it('returns all targeting for a feature',
+            (ctx) => {
+                expect(ctx.stdout).to.contain(JSON.stringify(mockTargeting, null, 2))
```
elliotCamblor, line 90: Do you have thoughts on JSON parsing the ctx.stdout rather than stringifying the expected value? It has the advantage of not needing to worry about the spacing, but there might be downsides I havent considered

laurawarr: I had considered doing it that way because it would make this `expect` more readable, but chose to go with this approach so that it _does_ validate the whitespace — meaning that we expect non-headless commands to output formatted json, so this would fail if that's not the case.

That being said, my next PR will change the targeting output to an [oclif tree](https://github.com/oclif/core/tree/a2efc040855969678a6aaf14d3ceb1d25d7a8411/src/cli-ux#uxtree) rather than JSON, so I'll likely need to switch this to some sort of snapshot anyway

elliotCamblor: Testing for the whitespace feels more like were testing the `writer.showResults` function over and over, but if its going to be a moot point soon anyway then its probably not even worth thinking about

elliotCamblor (approved):

PR #162, opened 2023-06-15
Use enquirer autocomplete for project select

emmawillis:
image

### DIFF
--- a/src/ui/promptForProject.ts
+++ b/src/ui/promptForProject.ts
@@ -1,18 +1,21 @@
-import inquirer from 'inquirer'
+import Enquirer from 'enquirer'
 import { Project } from '../api/schemas'
 
 export async function promptForProject(projects: Project[]): Promise<Project> {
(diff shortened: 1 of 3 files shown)


src/ui/promptForProject.ts
```diff
@@ -1,18 +1,21 @@
...
-            value: project
+            value: project._id
         }
     })
-    const responses = await inquirer.prompt([{
+    const responses = await new Enquirer<{project: string}>().prompt([{
         name: 'project',
         message: 'Which project do you want to use?',
-        type: 'list',
+        type: 'autocomplete',
```
laurawarr, line 16: we should add a TODO comment and/or create tickets to update the other list prompts to use autocomplete as well

emmawillis: done! https://taplytics.atlassian.net/browse/DVC-7835 i just put them all in the same ticket because the changes should be small now that it's working for one prompt

PR #163, opened 2023-06-15
fix: headless mode being ignored on variable and environment update

nsalamad:
# Changes
* this also addresses[ DVC-7620](https://taplytics.atlassian.net/browse/DVC-7620)
* move `populateParameters` into `Base` class
* add `key` validation for headless mode on variable and environment update
* display variable key along with the name on variable choice prompt

### DIFF
--- a/src/commands/base.ts
+++ b/src/commands/base.ts
@@ -5,7 +5,7 @@ import fs from 'fs'
 import path from 'path'
 import jsYaml from 'js-yaml'
 import { RepoConfigFromFile, UserConfigFromFile } from '../types'
-import { plainToClass } from 'class-transformer'
+import { ClassConstructor, plainToClass } from 'class-transformer'
 import { validateSync } from 'class-validator'
 import { reportValidationErrors } from '../utils/reportValidationErrors'
 import { getToken } from '../auth/getToken'
@@ -14,6 +14,8 @@ import { promptForProject } from '../ui/promptForProject'
 import inquirer from 'inquirer'
 import Writer from '../ui/writer'
 import { setDVCReferrer } from '../api/apiClient'
+import { Prompt } from '../ui/prompts'
+import { filterPrompts, mergeFlagsAndAnswers, validateParams } from '../utils/prompts'
 
 export default abstract class Base extends Command {
     static hidden = true
@@ -77,7 +79,7 @@ export default abstract class Base extends Command {
     authRequired = false
     // Override to true in commands that have "enhanced" functionality enabled by API access
     authSuggested = false
-    // Override to trye in commands that expect to run in the repo
+    // Override to true in commands that expect to run in the repo
     runsInRepo = false
 
     userConfig: UserConfigFromFile | null
@@ -227,5 +229,32 @@ export default abstract class Base extends Command {
 
     hasToken(): boolean {
         return this.authToken !== ''
+    }
+
+    public async populateParameters<ResourceType>(
+        paramClass: ClassConstructor<ResourceType>,
+        prompts: Prompt[],
+        flags: Record<string, unknown> = {},
+        isUpdate = false,
+    ): Promise<ResourceType> {
+        if (flags.headless) {
+            const params = plainToClass(paramClass, flags)
+            validateParams(paramClass, params, { whitelist: true, skipMissingProperties: isUpdate })
+            return params
+        }
+
+        const filteredPrompts = filterPrompts(prompts, flags)
+        const answers = await this.populateParametersWithInquirer(filteredPrompts)
+
+        const params = plainToClass(paramClass, mergeFlagsAndAnswers(flags, answers))
+        validateParams(paramClass, params, { whitelist: true, skipMissingProperties: isUpdate })
+        return params
+    }
+
+    protected async populateParametersWithInquirer(prompts: Prompt[]) {
+        return inquirer.prompt(prompts, {
+            token: this.authToken,
+            projectKey: this.projectKey
+        })
     }
 }
--- a/src/commands/environments/create.ts
+++ b/src/commands/environments/create.ts
@@ -12,7 +12,7 @@ import {
 } from '../../ui/prompts'
 import CreateCommand from '../createCommand'
 
-export default class CreateEnvironment extends CreateCommand<CreateEnvironmentParams> {
+export default class CreateEnvironment extends CreateCommand {
     static hidden = false
     static description = 'Create a new Environment for an existing Feature.'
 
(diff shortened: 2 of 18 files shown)


nsalamad:
* main
  * **PR #163** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/cli/163" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/cli/163?utm_source=stack-comment).

src/commands/base.ts
```diff
@@ -224,4 +225,84 @@ export default abstract class Base extends Command {
...
+            if (updatedFlags[key] === undefined) {
+                updatedFlags[key] = answers[key]
+            }
+        })
+        return updatedFlags
+    }
+
+    private async populateParametersWithInquirer(prompts: Prompt[], isUpdate = false) {
+        let filteredPrompts = prompts
+        if (isUpdate) {
```
laurawarr, line 280: rather than passing this parameter around, it could make sense to override this method in the UpdateCommand class that extends the Base class

src/commands/environments/create.ts
```diff
@@ -38,19 +38,22 @@ export default class CreateEnvironment extends CreateCommand<CreateEnvironmentPa
...
+        const params = await this.populateParameters(
+            CreateEnvironmentParams,
+            this.prompts,
+            {
+                key,
+                name,
+                description,
+                type,
+                headless
+            },
```
laurawarr, line 55: any reason to not just pass `flags` here?

src/commands/environments/update.ts
```diff
@@ -10,33 +11,68 @@ import {
...
+        const { key, name, type, description, headless } = flags
+
         await this.requireProject()
-        const { environment } = await inquirer.prompt<EnvironmentPromptResult>([environmentPrompt], {
-            token: this.authToken,
-            projectKey: this.projectKey
-        })
+        let envKey: string
+        if (headless) {
+            if (!key) {
```
laurawarr, line 42: To be consistent with the new commands we're adding, the key or ID to identify the environment to update should be passed as an argument.
For many resources a user could pass both a key to identify the resource while also modifying the key. To differentiate between the two, we're setting the pattern that args are used to identify the resource and flags are used to pass additional values (in this case the properties to update)

src/commands/environments/update.ts
```diff
@@ -10,33 +11,68 @@ import {
...
+                this.writer.showError('In headless mode, the key flag is required')
```
laurawarr, line 43: I know this was added in some other places, but I don't think "In headless mode" is necessary in this message. The key/ID is always required, it's just that in headless mode we don't prompt for it

src/ui/prompts/variablePrompts.ts
```diff
@@ -1,18 +1,25 @@
...
 
+export type VariablePromptResult = {
+    variable: VariableChoice['value']
+} & PromptResult
+
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const variableChoices = async (input: Record<string, any>):Promise<VariableChoice[]> => {
     const variables = await fetchVariables(input.token, input.projectKey)
     const choices = variables.map((variable) => {
+        const name = variable.name ? `${variable.name} ${chalk.gray(`(${variable.key})`)}` : variable.key
```
laurawarr, line 20: I believe we realized we need to use `dim` otherwise Jason can't see the logs 👀 
```suggestion
        const name = variable.name ? `${variable.name} ${chalk.dim(`(${variable.key})`)}` : variable.key
```

src/ui/prompts/variablePrompts.ts
```diff
@@ -1,18 +1,25 @@
...
         return {
-            name: variable.name || variable.key,
+            name: name,
```
laurawarr, line 22: nitpick: shorthand
```suggestion
            name,
```

src/commands/environments/create.ts
nsalamad: that's how it was before but i don't see a reason why we can't just pass in `flags`. Will make the change

src/ui/prompts/variablePrompts.ts
nsalamad: good catch!

src/commands/environments/create.ts
nsalamad: figured out why it was being done this way. [mergeFlagsAndAnswers](https://github.com/DevCycleHQ/cli/blob/main/src/commands/createCommand.ts#L69) was iterating over the `flags` argument's keys to set a value for a key that didn't get set from the users flags using the prompt answers. Passing in an object with those keys explicitly meant that even if `key`, `name`, `description`, `type` were `undefined`, they would still exist in the object so the merge code was able to set the missing values correctly. When I changed it just passing `flags` directly, the provided answers weren't set on the final result because those keys I mentioned above didn't exist on the object. 

Anyways, as a fix, I've updated `mergeFlagsAndAnswers` to [iterate](https://github.com/DevCycleHQ/cli/pull/163/files#diff-d62b22b357d8189368367ee647935585998107d28bfaf657bac900e1087a9597R274) over the keys of `answers`, this way whatever doesn't get set by the flags, _will_ get set by the answers from the prompts.

src/commands/variables/update.ts
```diff
@@ -21,23 +23,50 @@ export default class UpdateVariable extends UpdateCommand<CreateVariableParams>
...
+        await this.requireProject()
+        let variableKey: string
+        if (headless) {
+            if (!key) {
+                this.writer.showError('The key argument is required')
+                return
+            }
+            variableKey = key
+        } else {
+            const { variable } = await inquirer.prompt<VariablePromptResult>([variablePrompt], {
```
jsalaber, line 50: what happens when I put in a variable key that doesn't exist, won't this return nothing?

src/commands/base.ts
```diff
@@ -228,4 +229,60 @@ export default abstract class Base extends Command {
     hasToken(): boolean {
         return this.authToken !== ''
     }
+
+    public async populateParameters<ResourceType>(
```
jsalaber, line 233: is there any reason to even have these functions on the base command and not a util function? I was thinking of ripping out these methods since they don't really seem to use state from the Command, and it makes it harder to reuse these outside of the Command classes

src/commands/variables/update.ts
nsalamad: well in non headless mode you can't put in a variable key that doesn't exist because you are prompted to select from a set of keys in your project. But if you do headless mode and you put a key that doesn't exist, then we show an error that looks like `DevCycle Error: Variable with key {key} not found`. We're just logging the api response.

src/commands/base.ts
nsalamad: this method is generic enough that it can be used by most commands that require multiple argument/flags. It does use `authKey` and `projectKey` but yeah, I can see your point about it not accessing many of the base class properties directly. This wasn't really intended to be used outside of the Command classes - where are you trying to use it?

nsalamad: actually i could make `prompts` a member of the base class and then reference it in the methods like `this.prompts` - which would be nice because then i can drop another param from the method. I wish i could also do the same with `flags` but `flags` is static.

nsalamad: `filterPrompts` and `mergeFlagsAndAnswers` could def be util classes

jsalaber: `populateParameters` I feel like is doing too much and it makes it hard not to duplicate code that aren't in Commands (e.g. in the ListOption class). Just something to think about, I might tackle it in my PRs

jsalaber (approved):

PR #170, opened 2023-06-20
Use zod schemas for CreateEnvironmentParams and UpdateEnvironmentParams

emmawillis:
Note: i had to keep the old populateParameters method because it's still used by the other create/update functions, but it can be fully replaced by the zod version once they're all updated. i just figured it would be better to break this into smaller PRs and do one resource type at a time


src/commands/base.ts
```diff
@@ -251,6 +251,28 @@ export default abstract class Base extends Command {
...
+        if (!flags.headless) {
+            const filteredPrompts = filterPrompts(prompts, flags)
+            const answers = await this.populateParametersWithInquirer(filteredPrompts)
+            input = mergeFlagsAndAnswers(flags, answers)
+        }
+
+        const parse = schema.safeParse(input)
+
+        if (!parse.success) {
+            this.writer.showError(parse.error.message)
```
jsalaber, line 270: I think we should throw an error here, otherwise it's harder for functions calling this method to determine if there was an error or not

src/api/schemas.ts
```diff
@@ -7,3 +7,7 @@ export type SDKKeys = z.infer<typeof schemas.SDKKeys>
 export type Variable = z.infer<typeof schemas.Variable>
 export type Variation = z.infer<typeof schemas.Variation>
 export type Feature = z.infer<typeof schemas.Feature>
+export type CreateEnvironmentParams = z.infer<typeof schemas.CreateEnvironmentDto>
```
jsalaber, line 10: what are these for if they're being replaced by the dto?

jsalaber (approved):

jsalaber (changes requested):
oops meant to request changes

src/api/schemas.ts
emmawillis: the `CreateEnvironmentDto` in the schemas is a zodObject, so in order to use it as a typescript type we need to use `z.infer` to get the actual type of the parsed object. i tried a bunch of alternatives but couldn't find a better way to export both the type and the actual zodObject object that you can use for runtime validation, so i just have both here

src/commands/base.ts
emmawillis: good point, i changed it so `populateParametersWithZod` will throw and then the actual run methods for the commands can catch and handle the error

jsalaber (approved):

PR #171, opened 2023-06-20
Use zod schema for CreateFeatureDto

emmawillis:


jsalaber (approved):

src/commands/features/create.ts
```diff
@@ -56,6 +58,10 @@ export default class CreateFeature extends CreateCommand {
             const result = await createFeature(this.authToken, this.projectKey, params)
             this.writer.showResults(result)
         } catch (e) {
+            if (e instanceof ZodError) {
+                this.writer.showError(`Input failed validation with the following errors:\n${e.message}`)
```
jsalaber, line 62: you might actually need to change the test to account for this

emmawillis: ah yep just saw it failed - fixing now

PR #174, opened 2023-06-21
Use zod schema for CreateVariableDto

emmawillis:
swapped out CreateVariableParams for CreateVariableDto and UpdateVariableDto from the zod schema

also updated the error messages so that it shows `___ is a required field` if the field is missing and can also handle min and max length, and mismatched enums. 
image

unfortunately zod doesn't give you access to the failing regex if it's a regex issue, so the best i could do is log `format of ___ is invalid`
image

### DIFF
--- a/src/api/apiClient.ts
+++ b/src/api/apiClient.ts
@@ -1,6 +1,7 @@
 import axios, { AxiosError } from 'axios'
 import { BASE_URL } from './common'
 import { createApiClient } from './zodClient'
+import { ZodIssueCode, ZodIssueOptionalMessage, ErrorMapCtx } from 'zod'
 
 export const axiosClient = axios.create({
     baseURL: BASE_URL,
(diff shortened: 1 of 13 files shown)


src/api/apiClient.ts
```diff
@@ -35,5 +36,61 @@ axiosClient.interceptors.response.use(
...
+                return {
+                    message:
+                    `Invalid value: ${issue.path.join('.')} length must be >= ${issue.minimum} but got ${ctx.data}`,
+                }
+            } else {
+                return {
+                    message: 
+                    `Invalid value: ${issue.path.join('.')} must be >= ${issue.minimum} but got ${ctx.data}`,
+                }
+            }
```
laurawarr, line 58: just a nitpick, but we could condense this by using a ternary operator
```suggestion
           return {
               message: issue.type === 'string'
                   ? `Invalid value: ${issue.path.join('.')} length must be >= ${issue.minimum} but got ${ctx.data}`
                   : `Invalid value: ${issue.path.join('.')} must be >= ${issue.minimum} but got ${ctx.data}`
            }
```

src/api/apiClient.ts
```diff
@@ -35,5 +36,61 @@ axiosClient.interceptors.response.use(
...
+            } else {
+                return {
+                    message: 
+                    `Invalid value: ${issue.path.join('.')} must be <= ${issue.maximum} but got ${ctx.data}`,
+                }
+            }
+        case ZodIssueCode.invalid_string:
+            if (issue.validation === 'regex') {
+                return {
+                    message: `Invalid value: format of ${issue.path.join('.')} is invalid.`,
```
laurawarr, line 74: another nitpick, but considering all of these return an object with a message property prefixed by `Invalid value:` we could make a tiny util function within the errorMap to abstract that
```
(message: string) => ({ message: `Invalid value: ${message}` })
```

laurawarr (approved):
added a couple comments to reduce some of the duplication and improve readability in the error map, but otherwise looks good 👍

src/api/apiClient.ts
emmawillis: good point - fixed!

src/api/apiClient.ts
emmawillis: i just changed it to only print the 'Invalid value' if we don't know the details of the error and we're just blindly print the message, like:
`Invalid value: ${issue.path.join('.')}. ${issue.message || ''}`

but for something like `key length must be greater than X`, i'm thinking the 'invalid value' is a bit redundant?

PR #178, opened 2023-06-23
add tests for the variations create

elliotCamblor:


src/commands/variations/create.ts
```diff
@@ -59,39 +55,31 @@ export default class CreateVariation extends CreateCommand {
...
 
-        } catch (e) {
-            if (e instanceof ZodError) {
-                this.reportZodValidationErrors(e)
-                return
-            }
+        const variation = {
+            key: params.key,
+            name: params.name,
+            variables: variables ? JSON.parse(variables) : variableAnswers
```
nsalamad, line 79: I think we should wrap the `JSON.parse` call in a try/catch since someone might enter a malformed string

nsalamad (approved):

PR #179, opened 2023-06-23
Add features update command

emmawillis:
inquirer-autocomplete-prompt2

In order to be able to include the variables and sdkVisibility prompts in the update choices list and only prompt for them if they're selected, i had to make these changes:
- changed `sdkVisibilityPrompt` to `getSdkVisibilityPrompt` which returns a prompt we can include in the prompt lists instead of wrapping around the prompt function directly
- added `transformResponse` function to Prompt type to handle transforming the response from the prompt (because sdkVisibility has to be transformed from an array to an object)
- added `listOptionsPrompt` function to Prompt type to bypass calling inquirer.prompt directly and instead call a custom prompt wrapper (to trigger the VariationListOptions flow)
- created `handleCustomPrompts` function that decides whether to call `inquirer.prompt` or `listOptionsPrompt` for each prompt, and then transforms the response with transformResponse (this is called in `populateParametersWithInquirer`)

### DIFF
--- a/src/ui/prompts/commonPrompts.ts
+++ b/src/ui/prompts/commonPrompts.ts
@@ -1,5 +1,8 @@
 import chalk from 'chalk'
-import { Prompt } from './types'
+import { ListPrompt, Prompt } from './types'
+import { partition } from 'lodash'
+import inquirer from '../autocomplete'
+
 export const hintTextTransformer = (hint: string) =>
     (value: string, answers: unknown, { isFinal }: { isFinal: boolean }) => {
         let newValue = value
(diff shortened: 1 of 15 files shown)


src/ui/prompts/featurePrompts.ts
```diff
@@ -47,42 +46,37 @@ type SDKVisibilityChoice = {
...
-    sdkVisibility: SDKVisibilityChoice['value'][]
-}
 
 export const getSDKVisibilityChoices = (sdkVisibility?: Feature['sdkVisibility']): SDKVisibilityChoice[] => {
     return [
         {
             name: 'mobile',
             value: 'mobile',
-            checked: sdkVisibility?.mobile || true
+            checked: sdkVisibility?.mobile !== undefined ? sdkVisibility?.mobile : true
```
elliotCamblor, line 55: I think you should use `typeof 'undefined'` as the check instead in these cases

src/ui/prompts/commonPrompts.ts
```diff
@@ -36,3 +39,29 @@ export const descriptionPrompt: Prompt = {
...
+
+export const handleCustomPrompts = async (prompts: Prompt[], authToken: string, projectKey: string) => {
+    const [listOptionsPrompts, standardPrompts] = partition(prompts, (prompt) => !!prompt.listOptionsPrompt)
+
+    const result = await inquirer.prompt(standardPrompts, {
+        token: authToken,
+        projectKey: projectKey
+    })
+    for (const prompt of listOptionsPrompts) {
+        if (prompt.listOptionsPrompt) {
```
elliotCamblor, line 61: whats the reason for this check? Arent the prompts partitioned by the existence of this field?

emmawillis: good point, the problem was `partition` doesn't narrow the type so typescript was complaining that listOptionsPrompt could still be undefined here but i can get around it by casting it like: 
```
for (const prompt of listOptionsPrompts as ListPrompt[]) {
        result[prompt.name] = await prompt.listOptionsPrompt()
    }
```

elliotCamblor: yeah that seems fine to me

src/ui/prompts/featurePrompts.ts
emmawillis: good point, fixed!

laurawarr (approved):

PR #180, opened 2023-06-26
feat: add targeting update command

nsalamad:
# Changes
* Add `targeting update command`

https://github.com/DevCycleHQ/cli/assets/33500361/6840d103-df6c-48c8-a693-959ab7c41fa2

### DIFF
--- a/docs/targeting.md
+++ b/docs/targeting.md
@@ -6,6 +6,7 @@ Access and Modify Targeting Rules for a Feature with the Management API.
 * [`dvc targeting disable [FEATURE] [ENVIRONMENT]`](#dvc-targeting-disable-feature-environment)
 * [`dvc targeting enable [FEATURE] [ENVIRONMENT]`](#dvc-targeting-enable-feature-environment)
 * [`dvc targeting get [FEATURE] [ENVIRONMENT]`](#dvc-targeting-get-feature-environment)
+* [`dvc targeting update [FEATURE] [ENVIRONMENT]`](#dvc-targeting-update-feature-environment)
 
 ## `dvc targeting disable [FEATURE] [ENVIRONMENT]`
 
@@ -100,4 +101,36 @@ EXAMPLES
   $ dvc targeting get feature-one
 
   $ dvc targeting get feature-one environment-one
+```
+
+## `dvc targeting update [FEATURE] [ENVIRONMENT]`
+
+Update Targeting rules
+
+```
+USAGE
+  $ dvc targeting update [FEATURE] [ENVIRONMENT] [--config-path <value>] [--auth-path <value>] [--repo-config-path
+    <value>] [--client-id <value>] [--client-secret <value>] [--project <value>] [--no-api] [--headless] [--targets
+    <value>]
+
+ARGUMENTS
+  FEATURE      The Feature for the Targeting Rule.
+  ENVIRONMENT  The Environment where the Targeting Rule will be updated.
+
+FLAGS
+  --targets=<value>  List of targeting rules.
+
+GLOBAL FLAGS
+  --auth-path=<value>         Override the default location to look for an auth.yml file
+  --client-id=<value>         Client ID to use for DevCycle API Authorization
+  --client-secret=<value>     Client Secret to use for DevCycle API Authorization
+  --config-path=<value>       Override the default location to look for the user.yml file
+  --headless                  Disable all interactive flows and format output for easy parsing.
+  --no-api                    Disable API-based enhancements for commands where authorization is optional. Suppresses
+                              warnings about missing credentials.
+  --project=<value>           Project key to use for the DevCycle API requests
+  --repo-config-path=<value>  Override the default location to look for the repo config.yml file
+
+DESCRIPTION
+  Update Targeting rules
 ```
--- a/src/api/targeting.ts
+++ b/src/api/targeting.ts
@@ -1,9 +1,59 @@
+import { IsOptional, IsString, IsArray, IsNotEmpty } from 'class-validator'
 import apiClient from './apiClient'
+import { buildHeaders } from './common'
+import { Filter, UpdateFeatureConfigDto } from './schemas'
+
+export const filterTypes = [
+    'all',
+    'user',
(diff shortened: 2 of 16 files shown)


docs/targeting.md
```diff
@@ -101,3 +102,36 @@ EXAMPLES
...
+  $ dvc targeting update [FEATURE] [ENVIRONMENT] [--config-path <value>] [--auth-path <value>] [--repo-config-path
+    <value>] [--client-id <value>] [--client-secret <value>] [--project <value>] [--no-api] [--headless] [--name
+    <value>] [--targets <value>]
+
+ARGUMENTS
+  FEATURE      The Feature for the Targeting Rule.
+  ENVIRONMENT  The Environment where the Targeting Rule will be updated.
+
+FLAGS
+  --name=<value>     Human readable name
```
laurawarr, line 121: this may be incorrect, the feature config shouldn't have a name associated with it, only the individual targeting rules

nsalamad: yup you're right. Will remove

src/api/targeting.ts
```diff
@@ -1,11 +1,61 @@
+import { IsOptional, IsString, IsArray, IsNotEmpty } from 'class-validator'
```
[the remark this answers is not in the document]
laurawarr: oh I see, it just isn't acknowledging the decorators

src/commands/targeting/update.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    public async run(): Promise<void> {
+        const { args, flags } = await this.parse(UpdateTargeting)
+        const { feature, environment } = args
+        const { targets, headless } = flags
+
+        await this.requireProject()
+
+        if (headless && (!feature || !environment)) {
+            this.writer.showError('The key, feature and environment arguments are required')
```
laurawarr, line 45: doesn't look like a key is required here?
```suggestion
            this.writer.showError('The feature and environment arguments are required')
```

src/commands/targeting/update.ts
```diff
@@ -0,0 +1,120 @@
...
+        }),
+        environment: Args.string({
+            description: 'The Environment where the Targeting Rule will be updated.'
+        }),
+    }
+
+    static flags = {
+        ...UpdateCommand.flags,
+        targets: Flags.string({
+            description: 'List of trageting rules'
```
laurawarr, line 33:
```suggestion
            description: 'List of targeting rules'
```

laurawarr: We could possibly reference the API docs for the JSON schema expected here? https://docs.devcycle.com/management-api/#tag/Feature-Configurations/operation/FeatureConfigsController_update

src/ui/prompts/listPrompts/targetingListPrompt.ts
```diff
@@ -0,0 +1,97 @@
...
+        list[index] = {
+            name: target.audience.name,
+            value: target as UpdateTargetParams
+        }
+    }
+
+    transformToListOptions(list: UpdateTargetParams[]): ListOption<UpdateTargetParams>[] {
+        // TODO: default name is temporary until we have a better way to identify targeting rules
+        return list.map((target, index) => ({
+            name: target.audience.name || `Targeting Rule ${index + 1}`,
```
laurawarr, line 93: it'd be worth considering the target.name here as well, just in case the user defined it using the API

src/ui/prompts/listPrompts/targetingListPrompt.ts
```diff
@@ -0,0 +1,97 @@
...
+            return
+        }
+
+        const response = await inquirer.prompt([{
+            name: 'targetToEdit',
+            message: `Which ${this.itemType} would you like to edit?`,
+            type: 'list',
+            choices: list
+        }])
+        const index = list.findIndex((item) => item.name === response.targetToEdit.audience.name)
```
laurawarr, line 65: matching based on the name seems unreliable, since there's nothing preventing targets from having duplicate names

nsalamad: yeah, I agree. I don't really see a way to uniquely identify targets. I could stringify the entire target and match based on that value but even then there's nothing that stops someone from having the same defintion for a target

laurawarr: all the targets fetched from the API should have a unique `_id`

nsalamad: So, yeah I can use the `_id` property if it exists but what can I reliably use if doesn't? Newly added targets in the list won't have `_id`

elliotCamblor: The same thing just came up in Emma's PR. My suggestion was to pass along the `index` as part of the response object so you dont need to use `findIndex` to update the item. Think that might work?

src/commands/targeting/update.ts
nsalamad: it might be a little confusing to reference the API docs since we have different display names for properties. e.g. `name` references the audience name, `serve` is the `_variation` field of the distribution property, and definition is the `filters.filters` property of audience. If we reference the docs then we'd also have to explain the mapping of display names to actual properties. Also we might end up with a dead link in some versions of the CLI if the docs API URL ever changes. But I'm totally fine with putting it in if what I just said isn't that big of a deal

src/ui/prompts/listPrompts/targetingListPrompt.ts
nsalamad: Yeah I think that could work. But I'm gonna do it as apart of another ticket. The reason being that I'd need to change the definition for `transformToListOptions` which means I'd need to update all the other listPrompt classes. This PR is already big enough as it is

src/commands/targeting/update.ts
laurawarr: looks like this still needs to be updated

laurawarr (approved):
Added one comment about the key still being logged as required and we should create a ticket to document the targeting payload

PR #181, opened 2023-06-26
Prompt for variations in features create / update:

emmawillis:
- added the add/edit functions in VariationsListPrompt
- renamed `getVariationVariableValuePrompts` to `promptForVariationVariableValues` and created a new variationPrompt `getVariationVariablesPrompts` that just returns the prompts without calling inquirer.prompt (so we can use it in the prompts lists)
- added the option to set `previousReponseFields` on listOptions prompt types, which lets us pass previous responses along to the list fields. This way if you add or edit variables, the updated variables list is available in the variations prompt 

features create:
inquirer-autocomplete-prompt3

features update:
inquirer-autocomplete-prompt4

### DIFF
--- a/src/ui/prompts/listPrompts/variationsListPrompt.ts
+++ b/src/ui/prompts/listPrompts/variationsListPrompt.ts
@@ -1,5 +1,8 @@
 
-import { CreateVariationParams } from '../../../api/schemas'
+import { errorMap } from '../../../api/apiClient'
+import { CreateVariationDto, CreateVariationParams, Variable, Variation } from '../../../api/schemas'
+import inquirer from '../../autocomplete'
+import { getVariationVariablesPrompts, staticCreateVariationPrompts } from '../variationPrompts'
 import { ListOption, ListOptionsPrompt } from './listOptionsPrompt'
 import { AddItemPrompt, EditItemPrompt, RemoveItemPrompt, ContinuePrompt, ExitPrompt } from './promptOptions'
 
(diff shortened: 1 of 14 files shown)


src/ui/prompts/listPrompts/variationsListPrompt.ts
```diff
@@ -16,19 +19,74 @@ export class VariationListOptions extends ListOptionsPrompt<CreateVariationParam
...
+            message: 'Which variation would you like to edit?',
+            type: 'list',
+            choices: list
+        }])
+        const index = list.findIndex((listItem) => listItem.name === response.variationToEdit.name)
+
+        // Have a default for each of the prompts that correspond to the previous value of the variable
+        const filledOutPrompts = staticCreateVariationPrompts.map((prompt) => ({
+            ...prompt,
+            default: response.variationToEdit[prompt.name]
```
elliotCamblor, line 74: isn't `response.variationToEdit` an object in the form
```
{
	name: "string",
	value: variation
}
```

src/ui/prompts/listPrompts/variationsListPrompt.ts
```diff
@@ -16,19 +19,74 @@ export class VariationListOptions extends ListOptionsPrompt<CreateVariationParam
...
+            this.writer.warningMessage('No variations to edit')
+            return
+        }
+        const response = await inquirer.prompt([{
+            name: 'variationToEdit',
+            message: 'Which variation would you like to edit?',
+            type: 'list',
+            choices: list
+        }])
+        const index = list.findIndex((listItem) => listItem.name === response.variationToEdit.name)
```
elliotCamblor, line 69: This could potentially be problematic since I believe variation names are not necessarily unique. A potential workaround would be to include in the index as part of the response like so ```{
	name: "string",
	value: { variation, index }
}```
but id be interested to hear your thoughts

src/ui/prompts/listPrompts/variationsListPrompt.ts
emmawillis: hm it seems like response.variationToEdit just has the value field? the default values are working anyway

elliotCamblor: ah yeah, the answer is just the value field, forgot about that

src/ui/prompts/listPrompts/variationsListPrompt.ts
emmawillis: yeah our UX for 2 variations with the same name is pretty bad since we only show the name when you select from a list

we can at least make sure we update the right one by using key instead of name like 
      `const index = list.findIndex((listItem) => (listItem.value as Variation).key === response.variationToEdit.key)`
since we have access to the full variation in the list items 

(not sure why i can't get the type to work, since listItem.value should be a CreateVariationParams type, but it keeps insisting key doesn't exist)

emmawillis: oh wait never mind figured out the types issue, it wasn't passing the real `CreateVariationParams`, it was just an alias for the type passed to `promptEditItem`. I think I fixed it now

elliotCamblor: cool, that seems like a good option

elliotCamblor (approved):

PR #186, opened 2023-06-28
feat: add id field to ListOption

nsalamad:
# Changes
* add `id` field to the ListOption type so that subclasses that don't have a unique identifier can easily be identified
* add reordering prompt to targeting list
* fix reorder prompt

### DIFF
--- a/src/ui/prompts/listPrompts/listOptionsPrompt.ts
+++ b/src/ui/prompts/listPrompts/listOptionsPrompt.ts
@@ -9,7 +9,7 @@ import { AddItemPrompt, ContinuePrompt, EditItemPrompt, ExitPrompt, RemoveItemPr
  */
 export type ListOption<T> = {
     name: string
-    value: T
+    value: { item: T, id?: number }
 }
 
 export abstract class ListOptionsPrompt<T> {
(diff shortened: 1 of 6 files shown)


emmawillis (commented):
is it possible to add an error message like 'No targeting rules to reorder' if the list is empty? otherwise you get stuck on this screen with nothing to select

image

src/ui/prompts/listPrompts/listOptionsPrompt.ts
```diff
@@ -118,7 +118,9 @@ export abstract class ListOptionsPrompt<T> {
         try {
             switch (response) {
                 case 'add':
-                    newList.push(await this.promptAddItem())
+                    const newItem = await this.promptAddItem()
+                    newItem.value.id = newList.length
```
elliotCamblor, line 122: should this not be `newList.length - 1`? Or am I incorrectly assuming that the id field corresponds to the index?

nsalamad: yeah the id corresponds to the index but the item is being added to the list _after_ it's id is set so `newList.length` makes sense. First item gets 0, second item gets 1, third item gets 2... so on

nsalamad:
> is it possible to add an error message like 'No targeting rules to reorder' if the list is empty? otherwise you get stuck on this screen with nothing to select
> 
> image

good catch. Will Fix

src/ui/prompts/listPrompts/listOptionsPrompt.ts
elliotCamblor: ah true!

elliotCamblor (approved):

jsalaber (approved):

PR #188, opened 2023-06-28
chore: add AND to targeting tree

L-Joli:
- add AND to targeting tree

### DIFF
--- a/src/ui/targetingTree.ts
+++ b/src/ui/targetingTree.ts
@@ -71,9 +71,10 @@ const insertRulesTree = (rootTree: Tree, targets: FeatureConfig['targets'], vari
 
 const insertDefinitionTree = (rootTree: Tree, audience: Audience) => {
     const definitionTree = ux.tree()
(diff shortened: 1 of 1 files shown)


src/ui/targetingTree.ts
```diff
@@ -82,14 +82,18 @@ const insertDefinitionTree = (rootTree: Tree, audience: Audience) => {
             }
             userFilter.insert(comparatorMap[filter.comparator], values)
             const userProperty = filter.subType === 'customData' ? filter.dataKey : subTypeMap[filter.subType]
-            definitionTree.insert(userProperty, userFilter)
+            insertWithPrefixAND(definitionTree, userProperty, index, userFilter)
```
laurawarr, line 85: since the tree and value aren't relevant to the prefixing, we could just pass the string and index to this helper
```suggestion
            const prefixedProperty = prefixWithAnd(userProperty, index)
            definitionTree.insert(prefixedProperty, userFilter)
```

PR #189, opened 2023-06-29
feat: made features create command a quick create, added interactive mode

jsalaber:
# Summary

Made the default behaviour of `features create` to only prompt name, key, description and then use presets for `variables`, `variations`, and `targeting` to make it more seamlined.

### DIFF
--- a/src/commands/features/create.ts
+++ b/src/commands/features/create.ts
@@ -59,31 +59,44 @@ export default class CreateFeature extends CreateCommand {
         }
     }
     
-    async updateTargetingForQuickCreate(featureKey: string, developmentKey: string) {
-        await updateFeatureConfigForEnvironment(
-            this.authToken, 
-            this.projectKey, 
-            featureKey, 
-            developmentKey, {
-                targets: [{
-                    distribution: [{
-                        percentage: 1,
-                        _variation: 'variation-on',
-                    }],
-                    audience: {
-                        name: 'All Users',
-                        filters: {
-                            filters: [
-                                {
-                                    type: 'all'
-                                }
-                            ],
-                            operator: 'and'
+    async setupTargetingForEnvironments(featureKey: string) {
+        const environmentKeys = await this.getEnvironmentKeys()
+        Object.values(environmentKeys).forEach(async (environmentKey) => {
+            await updateFeatureConfigForEnvironment(
+                this.authToken, 
+                this.projectKey, 
+                featureKey, 
+                environmentKey, {
+                    targets: [{
+                        distribution: [{
+                            percentage: 1,
+                            _variation: 'variation-on',
+                        }],
+                        audience: {
+                            name: 'All Users',
+                            filters: {
+                                filters: [
+                                    {
+                                        type: 'all'
+                                    }
+                                ],
+                                operator: 'and'
+                            }
                         }
-                    }
-                }],
-                status: 'active'
-            })
+                    }],
+                    status: environmentKey === 'development' ? 'active' : 'inactive'
+                })
+        })
+    }
+
+    async getEnvironmentKeys() {
+        const configurations = await fetchEnvironments(this.authToken, this.projectKey)
+        const findEnvironmentKey = (type: string) => configurations.find((env) => env.type === type)?.key || type
+        return {
+            developmentKey: findEnvironmentKey('development'),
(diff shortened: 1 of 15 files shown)


jsalaber:
* main
  * **PR #189** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/cli/189" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/cli/189?utm_source=stack-comment).

src/commands/features/create.ts
```diff
@@ -21,9 +23,67 @@ export default class CreateFeature extends CreateCommand {
         sdkVisibility: Flags.string({
             description: 'The visibility of the feature for the SDKs'
         }),
+        interactive: Flags.boolean({
+            char: 'i',
+            description: 'Interactive Flag Creation Mode'
```
laurawarr, line 28: "Feature" would be more consistent with our terminology
```suggestion
            description: 'Interactive Feature Creation Mode'
```

src/commands/features/create.ts
```diff
@@ -21,9 +24,80 @@ export default class CreateFeature extends CreateCommand {
...
 
-    prompts = [keyPrompt, namePrompt, descriptionPrompt]
+    async getEnvironmentKeys() {
+        const configurations = await fetchEnvironments(this.authToken, this.projectKey)
+        const findEnvironmentKey = (type: string) => configurations.find((env) => env.type === type)?.key || type
+        return {
+            developmentKey: findEnvironmentKey('development'),
+            stagingKey: findEnvironmentKey('staging'),
+            productionKey: findEnvironmentKey('production'),
+        }
```
laurawarr, line 99: We should probably handle the case that there are multiple envs of each type. We could just create the config for every environment and then only enable the ones with type development

src/commands/features/create.ts
```diff
@@ -21,9 +24,80 @@ export default class CreateFeature extends CreateCommand {
...
+                            filters: {
+                                filters: [
+                                    {
+                                        type: 'all'
+                                    }
+                                ],
+                                operator: 'and'
+                            }
+                        }
+                    }],
```
laurawarr, line 86: it may make sense to pull these default configs into a separate file to reduce clutter in the command

src/commands/features/create.ts
```diff
@@ -33,14 +107,24 @@ export default class CreateFeature extends CreateCommand {
         if (headless && (!key || !name)) {
             this.writer.showError('The key and name flags are required')
             return
+        } 
+
+        if (!flags.interactive && !headless) {
+            const params = this.quickCreateFeatureParams(
```
laurawarr, line 113: are we allowing a way to use the quick create in headless mode?

src/ui/prompts/commonPrompts.ts
```diff
@@ -20,7 +20,12 @@ export const keyPrompt: Prompt = {
     name: 'key',
     message: 'Key',
     suffix: ':',
-    transformer: hintTextTransformer('(Unique ID)'),
+    default: (answers: Record<string, string>) => {
+        if (answers.name) {
+            return answers.name.trim().replace(/\s+/g, '-').toLowerCase()
```
nsalamad, line 25: during an update, don't we want to set the default as the current key since the name in kebab case might be different than the key?

jsalaber: during an update it won't have the answer for the name as it won't have been prompted in succession of a name prompt

nsalamad: the order of prompting has been swapped for features updates and variations update. It goes name and then key. But also, shouldn't we be returning the current key if it exists as the default instead of an empty string

jsalaber: in `feature update` you choose to select the fields you're updating, name isn't guaranteed to be prompted if you're updating key. Although if you choose both name and key it will infer the key from the new name you set

nsalamad: ahhh I see. I guess it makes sense to default the key to the new name if you're trying to update both key and name. 

Oh, I thought we were already defaulting to the previous value.

nsalamad (approved):

PR #191, opened 2023-06-29
chore: add colors to the tree

L-Joli:
- add colors to the tree
green
red

### DIFF
--- a/src/ui/constants/colors.ts
+++ b/src/ui/constants/colors.ts
@@ -1,0 +1,6 @@
+export const COLORS = {
+    lightBlue: '#93E3FE',
+    lightGreen: '#B0DC8B',
+    lightYellow: '#FDF894',
+    coral: '#D97575'
+}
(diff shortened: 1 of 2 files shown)


src/ui/targetingTree.ts
```diff
@@ -156,4 +165,16 @@ const insertScheduleTree = (rootTree: Tree, rollout?: Rollout) => {
 
         rootTree.insert('gradual rollout', scheduleTree)
     }
+}
+
+const coloredTitle = (title: string) => {
+    return chalk.hex('#93E3FE')(title)
```
laurawarr, line 171: it seems like it'd be worthwhile defining these hex codes as constants, maybe in the UI folder? then we can easily reuse these colours

src/ui/constants/colors.ts
```diff
@@ -0,0 +1,6 @@
+export const COLORS = {
+    lightBlue: '#93E3FE',
+    paleGreen: '#B0DC8B',
+    paleYellow: '#FDF894',
```
laurawarr, line 4: just a naming nitpick, but why not stick with either "pale" or "light"?

laurawarr (approved):
one comment, but otherwise looks good 👍

src/ui/constants/colors.ts
L-Joli: 😂 I'm usually bad at choosing names, so I followed chatGPT suggestions for the color names, but you're right, let's stick to 'light'🤣

PR #193, opened 2023-06-29
Dvc 8010 log command suggestion

L-Joli:
- add command suggestion to enable the feature if it's disabled
command log


src/commands/targeting/__snapshots__/update.test.ts.snap
```diff
@@ -59,10 +59,24 @@ All Users New
...
 
 exports[`Targeting update updates a target in headless mode 1`] = `
 "{\\"_feature\\":\\"647e36gg16d9k4815fi3f97d\\",\\"_environment\\":\\"648b421gg3f682869c0f22f8\\",\\"_createdBy\\":\\"test\\",\\"status\\":\\"inactive\\",\\"updatedAt\\":\\"2023-06-27T03:19:58.541Z\\",\\"targets\\":[{\\"_id\\":\\"647e36gg16d9k2814fc3f15e\\",\\"audience\\":{\\"name\\":\\"All Users\\",\\"filters\\":{\\"operator\\":\\"and\\",\\"filters\\":[{\\"type\\":\\"all\\"}]}},\\"distribution\\":[{\\"_variation\\":\\"647e36gg16c1d4814fe3f962\\",\\"percentage\\":1}]}],\\"readonly\\":false}
+
+Targeting rules updated.
+
+To enable this feature on this environment, use:
+
+dvc targeting enable feature-key env-key
+
```
laurawarr, line 80: We shouldn't output this hint in headless mode — in headless mode users may want to parse the json output to use elsewhere

src/commands/targeting/update.ts
```diff
@@ -117,5 +121,18 @@ export default class UpdateTargeting extends Base {
...
         this.writer.showResults(result)
+        this.showSuggestedCommand(featureKey, envKey, result, isTargetingUpdated)
+    }
+
+    private showSuggestedCommand(featureKey: string, envKey: string, result: FeatureConfig, isTargetingUpdated: boolean = true) {
+        const { status, targets } = result
+        if (status === 'active' || targets.length === 0)  {
+            return
+        }
+        const message = (isTargetingUpdated ? '\nTargeting rules updated.\n' : '') +
```
laurawarr, line 132: It seems a bit excessive to do a deep comparison of the feature configs just to print this message. Considering updating targeting rules is the only thing this command does, and at this point the request has been successfully sent whether the payload changed or not, could we either always show this or just omit it entirely?

L-Joli: hmm...should we just omit it then? caz I was trying not to show this line for continue and exit prompt, don't want it be a misleading message. and you're right, I feel like the comparison is excessive too...

laurawarr (approved):

PR #196, opened 2023-06-30
chore: add validation to prompts

nsalamad:
# Changes
* Add validation for required values, kebob case, and max length
* Fix bug with listPromptOption when discarding changes
* Fix bug where `variation`/`variables` was being sent as an empty array after discarding changes for variation/variables changes on features update

### DIFF
--- a/src/utils/validators.ts
+++ b/src/utils/validators.ts
@@ -1,0 +1,20 @@
+export const isRequired = (inputName: string, value: string) => {
+    if (!value || value.trim().length === 0) {
+        return `${inputName} is required.`
+    }
+    return true
+}
+
+export const isValidKey = (inputName: string, value: string) => {
+    if (!/^[\w-.]+$/.test(value)) {
+        return `${inputName} can only contain letters, dashes, dots and underscores.`
+    }
+    return true
+}
+
+export const maxLength = (inputName: string, value: string, max: number) => {
+    if (value.length > max) {
+        return `${inputName} cannot be longer than ${max} characters.`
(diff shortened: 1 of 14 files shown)


laurawarr (approved):

src/utils/validators.ts
```diff
@@ -0,0 +1,20 @@
+export const isRequired = (inputName: string, value: string) => {
+    if (!value || value.length === 0) {
+        return `${inputName} is required.`
+    }
+    return true
+}
+
+export const isKebabCase = (inputName: string, value: string) => {
+    if (!/^[\w-.]+$/.test(value)) {
+        return `${inputName} can only contain letters, dashes, dots and underscores.`
```
laurawarr, line 10: just a naming nitpick, this seems to be validating more than just kebab case, would something like "isValidKey" be more accurate?

src/utils/validators.ts
```diff
@@ -0,0 +1,20 @@
+export const isRequired = (inputName: string, value: string) => {
+    if (!value || value.length === 0) {
```
laurawarr, line 2: may be worth adding a trim to make sure it's not just whitespace characters?
```suggestion
    if (!value || value.trim().length === 0) {
```

laurawarr (approved):
A couple small comments on the validation helpers, but otherwise looks good

src/utils/validators.ts
nsalamad: agreed. Will make the change

PR #199, opened 2023-06-30
variations update should now properly accept feature keys as an arg

elliotCamblor:


src/commands/variations/update.ts
```diff
@@ -65,11 +66,14 @@ export default class UpdateVariation extends UpdateCommand {
         } else {
             selectedVariation = await fetchVariationByKey(this.authToken, this.projectKey, featureKey, args.key)
         }
-
+        const feature = await fetchFeatureByKey(this.authToken, this.projectKey, featureKey)
```
laurawarr, line 69: could we just do this if we're not using the feature prompt? otherwise we're fetching the feature from the api twice

elliotCamblor: I still need to fetch even if were not using the feature prompt, since the feature prompt doesnt return the full feature. I made a ticket to refactor that because the prompt is being used in several different places

laurawarr: Won't the request to fetch the variables associated with the feature already return an error like this if the key is invalid?

elliotCamblor: yeah I think youre right. Pretty sure I got baited by typescript here

src/commands/variations/update.ts
```diff
@@ -65,11 +66,11 @@ export default class UpdateVariation extends UpdateCommand {
         } else {
             selectedVariation = await fetchVariationByKey(this.authToken, this.projectKey, featureKey, args.key)
         }
-
+        const feature = await fetchFeatureByKey(this.authToken, this.projectKey, featureKey) as Feature
         this.prompts.push(await getVariationVariablePrompt(
             this.authToken,
             this.projectKey,
-            featureKey,
+            feature._id,
```
laurawarr, line 73: this seems like it would be an API bug, all the endpoints should accept either a key or an ID 🤔

laurawarr (approved):

src/commands/variations/update.ts
elliotCamblor: Even for query params? This field in the query params class called `isObjectIDorFalse` which makes it seem intentional

elliotCamblor: Im just not sure why 😅

laurawarr: Ideally all values could be accepted as keys, I'm assuming we made this choice to avoid querying the feature by key just to build the variable query params. We could consider updating the API, but that's out of scope of this work 😛

PR #203, opened 2023-07-04
chore: update description to explain definition and serve for targeting update

jsalaber:


jsalaber:
* main
  * **PR #203** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/cli/203" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/cli/203?utm_source=stack-comment).

src/commands/targeting/update.ts
```diff
@@ -18,7 +18,8 @@ import { FeatureConfig } from '../../api/schemas'
 
 export default class UpdateTargeting extends Base {
     static hidden = false
-    static description = 'Update Targeting rules'
+    // eslint-disable-next-line max-len
+    static description = 'Update Targeting rules for a Feature. Definition is the audience for the feature, while serve is the variation to serve to the audience.'
```
nsalamad, line 22: would it be possible to surround 'definition' and 'serve' backticks(`)? Also i think we should be specific about the fact that `serve` should be the key or the mongo ID of the variation to serve

jsalaber: it won't format in the console with backticks unfortunately

nsalamad: oh sorry, i meant for the markdown file. Don't remember if the markdown text is generated from the string here

nsalamad (approved):
If you're able to add backticks, cool. If not, all good

PR #210, opened 2023-07-05
Dvc 8076 refactor feature prompt

elliotCamblor:
- Change FeaturePrompt to return the full feature object rather than just the key
- VariableFeaturePrompt still just returns the key, uses a transformer on the feature prompt answer to do so
- Update everywhere that uses FeaturePrompt to support the new output
- Remove a bunch of redundant API calls to fetch the feature
- Fix a bug with variations create where it would send the featureKey as a query param to variables GET
- Update all affected unit tests to mock out the correct api calls

### DIFF
--- a/src/commands/features/delete.ts
+++ b/src/commands/features/delete.ts
@@ -2,7 +2,7 @@ import { Args } from '@oclif/core'
 import { deleteFeature } from '../../api/features'
 import Base from '../base'
 import inquirer from 'inquirer'
-import {featurePrompt, FeaturePromptResult} from '../../ui/prompts'
+import { featurePrompt, FeaturePromptResult } from '../../ui/prompts'
 
 export default class DeleteFeatures extends Base {
     static hidden = false
--- a/src/commands/features/delete.ts
+++ b/src/commands/features/delete.ts
@@ -22,11 +22,11 @@ export default class DeleteFeatures extends Base {
 
         let featureKey
         if (!args.feature) {
-            const { feature  } = await inquirer.prompt([featurePrompt], {
+            const { feature } = await inquirer.prompt([featurePrompt], {
                 token: this.authToken,
                 projectKey: this.projectKey
             })
-            featureKey = feature
+            featureKey = feature.key
         } else {
             featureKey = args.feature
         }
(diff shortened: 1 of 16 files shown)


src/commands/variations/create.ts
```diff
@@ -43,13 +44,14 @@ export default class CreateVariation extends CreateCommand {
...
+        let feature_id
         if (!args.feature) {
             const { feature } = await inquirer.prompt([featurePrompt], {
                 token: this.authToken,
                 projectKey: this.projectKey
             })
-
-            featureKey = feature
+            featureKey = feature.key
+            feature_id = feature._id
```
nsalamad, line 54: nitpick: should this be in camel case?

elliotCamblor: normally we use snake case for fields that contain mongo ids, but im open to debate on it

elliotCamblor: Though it does have precedent in this repo already to use snake case

jsalaber (approved):

src/commands/variations/create.ts
nsalamad: ah i see, nvm then

src/commands/features/delete.ts
```diff
@@ -22,11 +22,11 @@ export default class DeleteFeatures extends Base {
 
         let featureKey
         if (!args.feature) {
-            const { feature  } = await inquirer.prompt([featurePrompt], {
+            const { feature } = await inquirer.prompt([featurePrompt], {
```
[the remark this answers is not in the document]
elliotCamblor: maybe thats possible with inquirer? That would be its own ticket imo since we would need to update all of the prompts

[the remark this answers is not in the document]
jsalaber: we already do it in a number of places in the code, for example in the update variables code we use it to explicitly type the result of the prompt: 
```
const { variable } = await inquirer.prompt<VariablePromptResult>([variablePrompt], {
```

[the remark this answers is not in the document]
elliotCamblor: idk how I hadn't noticed that! Ill add it to some of the prompts I changed in this PR

PR #211, opened 2023-07-05
chore: output targeting enable disable as a tree

jsalaber:
# Changes

- Refactor enable / disable targeting commands to use common method
- output response of enable / disable targeting as a tree

### DIFF
--- a/src/commands/targeting/disable.ts
+++ b/src/commands/targeting/disable.ts
@@ -1,8 +1,10 @@
 import { Args } from '@oclif/core'
-import inquirer from '../../ui/autocomplete'
 import { disableTargeting } from '../../api/targeting'
-import { EnvironmentPromptResult, environmentPrompt, featurePrompt } from '../../ui/prompts'
 import Base from '../base'
+import { fetchEnvironmentByKey, fetchEnvironments } from '../../api/environments'
+import { fetchVariations } from '../../api/variations'
+import { renderTargetingTree } from '../../ui/targetingTree'
+import { getFeatureAndEnvironmentKeyFromArgs } from '../../utils/targeting'
 
 export default class DisableTargeting extends Base {
     static hidden = false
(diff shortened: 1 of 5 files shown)


jsalaber:
* main
  * **PR #211** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/cli/211" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/cli/211?utm_source=stack-comment).

src/utils/targeting/index.ts
```diff
@@ -0,0 +1,46 @@
+import { disableTargeting, enableTargeting } from '../../api/targeting'
+import inquirer from '../../ui/autocomplete'
+import { featurePrompt, EnvironmentPromptResult, environmentPrompt } from '../../ui/prompts'
+
+export const getResponsesForEnableDisableTargeting = async (
```
laurawarr, line 5: Would this be more accurate as something like `getFeatureAndEnvironmentArgs`? I had to read the function code to understand what "responses" this is getting. I could also see us using this for targeting commands other than enable/disable

src/commands/targeting/enable.ts
```diff
@@ -23,47 +28,26 @@ export default class EnableTargeting extends Base {
...
-            responses.featureKey,
-            responses.environmentKey
-        )
-        return this.writer.showResults(enableTargetingForFeatureAndEnvironment)
+            responses.featureKey as string,
+            responses.environmentKey as string
+        ) 
+    
+        if (flags.headless) {
+            this.writer.showResults(getResponsesForEnableDisableTargeting)
```
laurawarr, line 45: shouldn't this take `updatedTargeting` rather than `getResponsesForEnableDisableTargeting`?

jsalaber: oops yeah

src/commands/targeting/disable.ts
```diff
@@ -23,47 +25,26 @@ export default class DisableTargeting extends Base {
...
-        return this.writer.showResults(disableTargetingForFeatureAndEnvironment)
+            responses.featureKey as string,
+            responses.environmentKey as string
+        ) 
+    
+        if (flags.headless) {
+            this.writer.showResults(updatedTargeting)
+        } else {
+            // TODO: reuse the data fetched for the prompts
+            const environments = await fetchEnvironments(this.authToken, this.projectKey)
```
laurawarr, line 45: we could just fetch the related environment by key, since we're only updating one

laurawarr (approved):
one small comment regarding fetch the env, but otherwise looks good

PR #212, opened 2023-07-06
chore: handle outputting in targeting tree audience match filters

jsalaber:
# Changes

- handle audience match filter when rendering targeting list
- added audiences to constructor of list option classes to prevent fetching in multiple places

### DIFF
--- a/src/commands/targeting/disable.ts
+++ b/src/commands/targeting/disable.ts
@@ -4,6 +4,7 @@ import { Variation } from '../../api/schemas'
 import Base from '../base'
 import { renderTargetingTree } from '../../ui/targetingTree'
 import { getFeatureAndEnvironmentKeyFromArgs } from '../../utils/targeting'
+import { fetchAudiences } from '../../api/audiences'
 
 export default class DisableTargeting extends Base {
     static hidden = false
@@ -45,10 +46,12 @@ export default class DisableTargeting extends Base {
         if (flags.headless) {
             this.writer.showResults(updatedTargeting)
         } else {
+            const audiences = await fetchAudiences(this.authToken, this.projectKey)
             renderTargetingTree(
-                [updatedTargeting],
+                [updatedTargeting], 
                 [environment],
-                feature.variations as Variation[]
+                feature.variations as Variation[],
+                audiences
             )
         }
     }
--- a/src/commands/targeting/enable.ts
+++ b/src/commands/targeting/enable.ts
@@ -4,6 +4,7 @@ import { Variation } from '../../api/schemas'
 import { renderTargetingTree } from '../../ui/targetingTree'
 import Base from '../base'
 import { getFeatureAndEnvironmentKeyFromArgs } from '../../utils/targeting'
+import { fetchAudiences } from '../../api/audiences'
 
 export default class EnableTargeting extends Base {
     static hidden = false
@@ -45,10 +46,12 @@ export default class EnableTargeting extends Base {
         if (flags.headless) {
             this.writer.showResults(updatedTargeting)
         } else {
+            const audiences = await fetchAudiences(this.authToken, this.projectKey)
             renderTargetingTree(
-                [updatedTargeting],
+                [updatedTargeting], 
                 [environment],
-                feature.variations as Variation[]
+                feature.variations as Variation[],
+                audiences
             )
         }
     }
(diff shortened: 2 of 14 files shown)


src/utils/audiences/index.ts
```diff
@@ -0,0 +1,36 @@
...
+    newTarget.audience.filters.filters = filters
+    return newTarget
+}
+
+export const replaceAudienceIdInFilter = (filter: Filter, audiences: Audience[]) => {
+    const newFilter = structuredClone(filter) as Filter
+    if (newFilter.type === 'audienceMatch') {
+        const audienceIds = newFilter._audiences!
+        const audienceNames = audienceIds.map((audienceId) => {
+            const audience = audiences.find((audience) => audience._id === audienceId)
```
laurawarr, line 30: it may make sense to build a map of audience IDs to audience names so that we don't need to search the list for each filter, considering we're working with a list of all audiences for the project

laurawarr: we could also save ourselves from iterating the targets and filters again here if we just pass the audiences as an optional parameter to the tree builder, then it can check the value in the map if it encounters an audience match filter

src/commands/targeting/enable.ts
```diff
@@ -45,10 +47,12 @@ export default class EnableTargeting extends Base {
         if (flags.headless) {
             this.writer.showResults(updatedTargeting)
```
nsalamad, line 48: not related to this PR but do you know why we're not using the tree view in headless mode? we have all the required info to do it

src/commands/targeting/disable.ts
```diff
@@ -4,6 +4,8 @@ import { Variation } from '../../api/schemas'
 import Base from '../base'
 import { renderTargetingTree } from '../../ui/targetingTree'
 import { getFeatureAndEnvironmentKeyFromArgs } from '../../utils/targeting'
+import { fetchAudiences } from '../../api/audiences'
+import { replaceAudienceIdsInFeatureConfigs } from '../../utils/audiences'
```
laurawarr, line 8: looks like this import is no longer used?

jsalaber: I'll clean up the imports

src/commands/targeting/enable.ts
nsalamad: ah okay

src/ui/targetingTree.ts
```diff
@@ -36,11 +37,18 @@ const comparatorMap = {
...
     const targetingTree = ux.tree()
-
-    featureConfigs.forEach((config) => {
+    let configs = featureConfigs
+    // if there's audience match filter, we want to build audience name map
+    // and replace audience ids in feature configs with audience names
+    if (configs.some((config) => hasAudienceMatchFilter(config))) {
+        const audienceNameMap = buildAudienceNameMap(audiences)
+        configs = replaceAudienceIdsInFeatureConfigs(configs, audienceNameMap)
+    }
```
laurawarr, line 50: could we not save the iteration of checking for an audience match filter and replacing the audience ids in the feature config by just building and passing the audience map to `insertRulesTree`? since it's already handling replacing the values if they exist

jsalaber: I tried it out that way and it felt cumbersome having another a function argument be passed on to multiple levels of nested calls, all while being optional. 

This way the configs get changed at the top and the nested functions stay untouched

laurawarr: Ah, it looks like `insertRulesTree` still has the param but it's not being used. The main advantage I'd see is that we'd avoid having `replaceAudienceIdsInFeatureConfigs` duplicating the logic of traversing the feature config

jsalaber: I think it would be easier if all of these methods were encapsulated in some sort of class instead, as it would help prevent then need to pass around these parameters in such a nested way and have it available on the instance. I can make a ticket for it, as I would argue the duplication of traversing is worth it for now to decrease the amount of arguments we need to pass down

laurawarr: I agree that a class instance would be a better way to store a reference to the associated envs/variations/audiences

laurawarr: I still think this is traversing the config two more times than it needs to. I'd advocate we pass the param to `insertRulesTree` and then create a follow up ticket to convert this to a class

PR #214, opened 2023-07-07
chore: make targeting update command behave the same as other update commands

nsalamad:
# Changes
* Update `targeting update` test
* Change `targeting update` command to use `populateParametersWithZod`
* Can now also update the env status from the `targeting update` command

### DIFF
--- a/docs/targeting.md
+++ b/docs/targeting.md
@@ -111,13 +111,15 @@ Update Targeting rules for a Feature. The definition is the audience for the fea
 USAGE
   $ dvc targeting update [FEATURE] [ENVIRONMENT] [--config-path <value>] [--auth-path <value>] [--repo-config-path
     <value>] [--client-id <value>] [--client-secret <value>] [--project <value>] [--no-api] [--headless] [--targets
-    <value>]
+    <value>] [--status enable|disable]
 
 ARGUMENTS
   FEATURE      The Feature for the Targeting Rule.
   ENVIRONMENT  The Environment where the Targeting Rule will be updated.
 
 FLAGS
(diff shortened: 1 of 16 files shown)


src/commands/targeting/__snapshots__/update.test.ts.snap
```diff
@@ -92,6 +92,19 @@ To enable this feature on this environment, use:
 `;
 
 exports[`targeting update updates a target in headless mode 1`] = `
-"{\\"_feature\\":\\"647e36gg16d9k4815fi3f97d\\",\\"_environment\\":\\"648b421gg3f682869c0f22f8\\",\\"_createdBy\\":\\"test\\",\\"status\\":\\"inactive\\",\\"updatedAt\\":\\"2023-06-27T03:19:58.541Z\\",\\"targets\\":[{\\"_id\\":\\"647e36gg16d9k2814fc3f15e\\",\\"audience\\":{\\"name\\":\\"All Users\\",\\"filters\\":{\\"operator\\":\\"and\\",\\"filters\\":[{\\"type\\":\\"all\\"}]}},\\"distribution\\":[{\\"_variation\\":\\"647e36gg16c1d4814fe3f962\\",\\"percentage\\":1}]}],\\"readonly\\":false}
+"└─ test-env
```
elliotCamblor: is it expected that commands output as a tree in headless mode? I thought that was not the case

nsalamad: you're right, it should not be output as a tree in headless mode. Already reverted this change

elliotCamblor (approved):

PR #219, opened 2023-07-11
feat: add release yaml

L-Joli:
- add release workflow


.github/workflows/release.yml
```diff
@@ -0,0 +1,129 @@
...
+
+jobs:
+
+  checks:
+    runs-on: ubuntu-latest
+    outputs:
+      HAS_TAG: ${{ steps.check_tag.outputs.has_tag }}
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
```
JamieSinn, line 16: You're using git describe tags - so you'll need to add checkout depth 0 to this.

.github/workflows/release.yml
```diff
@@ -0,0 +1,129 @@
...
+      
+      - name: Check if commit has a new tag
```
JamieSinn, line 18: You can change this to be a trigger on a tag. Instead of having it be a check against tags

.github/workflows/release.yml
```diff
@@ -0,0 +1,129 @@
...
+            echo "has_tag=false" >> $GITHUB_OUTPUT
+            echo false
+          fi
+
+  release:
+    runs-on: ubuntu-latest
+    needs: checks
+    outputs:
+      CLI_VERSION: ${{ steps.get_cli_version.outputs.CLI_VERSION }}
+    if: needs.checks.outputs.HAS_TAG == 'false'
```
JamieSinn, line 34: This can be a trigger condition instead. Check out the docs:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore

.github/workflows/release.yml
```diff
@@ -0,0 +1,129 @@
...
+    steps:      
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - name: Set Git author
+        run: |
+          git config user.name 'github-actions[bot]'
+          git config user.email 'github-actions[bot]@users.noreply.github.com'
```
JamieSinn, line 42: This isn't needed - it's auto configured

.github/workflows/release.yml
```diff
@@ -0,0 +1,129 @@
...
+      
+      - name: Get node version
+        id: get_node_version
+        run: |
+          echo "NVMRC=$(cat .nvmrc)" >> $GITHUB_OUTPUT
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: "${{ steps.get_node_version.outputs.NVMRC }}"
```
JamieSinn, line 52: This is an interesting way of doing it - but NVM is already installed by default: https://github.com/actions/runner-images/issues/4 and you can use it by default

.github/workflows/release.yml
```diff
@@ -0,0 +1,129 @@
...
+        
+      - name: Build
+        run: yarn build
+
+      - name: Bump CLI version
+        run: npm version patch --force
+
+      - name: Read version from package.json
+        id: get_cli_version
+        run: echo "CLI_VERSION=$(node -p "require('./package.json').version")" >> $GITHUB_OUTPUT
```
JamieSinn, line 65: It might be easier to use JQ here to pull the version from the JSON itself.

.github/workflows/release.yml
```diff
@@ -0,0 +1,129 @@
...
+
+      - name: Get latest tag
+        id: get_latest_tag
+        run: echo "LATEST_TAG=$(git describe --abbrev=0 --tags)" >> $GITHUB_OUTPUT
```
JamieSinn, line 69: Because you're using git describe - you'll need the checkout depth 0 in the checkout step to include tags.

JamieSinn: Alternatively - you can use a pre-made action for this

.github/workflows/release.yml
```diff
@@ -0,0 +1,129 @@
...
+          git tag -f ${{ steps.get_latest_tag.outputs.LATEST_TAG }}
+          git push origin main --follow-tags
+      
+      - name: Create GitHub Release
+        uses: ncipollo/release-action@v1
+        with:
+          name: "v${{ steps.get_cli_version.outputs.CLI_VERSION }}"
+          tag: "v${{ steps.get_cli_version.outputs.CLI_VERSION }}"
+          generateReleaseNotes: "true"
+          makeLatest: "true"
```
JamieSinn, line 84: This should be replaced with https://github.com/devcyclehq/release-action

.github/workflows/release.yml
```diff
@@ -0,0 +1,129 @@
...
+          
+      - name: Publish to NPM
+        run: npm publish --access public
```
JamieSinn, line 87: You'll need to use an NPM Auth token for this - as part of the environment with the step run.

.github/workflows/release.yml
```diff
@@ -0,0 +1,129 @@
...
+          
+      - name: Update CLI version in docs repo
+        run: |
+          ls -al
+          echo $BRANCH_NAME
+          git checkout -b "$BRANCH_NAME"
+          git branch
+          sed -i 's/const DVC_CLI_VERSION = .*/const DVC_CLI_VERSION = '\''v${{ needs.release.outputs.CLI_VERSION }}'\'' \/\/ auto updated by dvc cli release workflow/' docusaurus.config.js
+          git add docusaurus.config.js
+          git commit -m "Update CLI version to v${{ needs.release.outputs.CLI_VERSION }}"
```
JamieSinn, line 118: This should be a shell script that's parameterized to make it more manageable - and the outputs should be cleaned up to make sure that it's not messy in the output on GHA.

JamieSinn (changes requested):

.github/workflows/release.yml
L-Joli: ohh cool!

.github/workflows/release.yml
```diff
@@ -0,0 +1,22 @@
...
+  push:
+    tags:
+      - '*'
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Create GitHub Release
+        uses: DevCycleHQ/release-action@main
```
JamieSinn, line 13: This needs to include the full path `gh-release`

.github/workflows/cli-release.yml
```diff
@@ -0,0 +1,83 @@
...
+        type: boolean
+        default: true
+      npm-version:
+        description: 'NPM version bump?'
+        required: false
+        type: choice
+        options: 
+          - major
+          - minor
+          - patch
```
rob-odwyer, line 23: This choice input is neat, I'm stealing this for the other release workflows 😄

.github/actions/release/action.yml
```diff
@@ -0,0 +1,47 @@
...
+runs:
+  using: 'composite'
+  steps:
+    # - name: Create GitHub Release
+    #   uses: DevCycleHQ/release-action/gh-release@main
+    #   with:
+    #     github-token: ${{ inputs.github-token }}
+    #     draft: ${{ inputs.draft }}
+    #     prerelease: ${{ inputs.prerelease }}
+    #     tag: # may need tag here
```
rob-odwyer, line 29: We're going to be making some breaking changes to DevCycleHQ/release-action pretty soon, so I'd say it's not worth keeping a commented version around.

.github/workflows/cli-release.yml
```diff
@@ -0,0 +1,83 @@
...
+  contents: write
+
+jobs:
+  cli-release:
+    runs-on: ubuntu-latest
+    steps:      
+      - name: Check out code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
rob-odwyer, line 38: I think I needed to add this option alongside `fetch-depth` to have permission to push to the main branch from the action:
```
token: ${{ secrets.AUTOMATION_USER_TOKEN }}
```
That secret is defined on the DevCycleHQ organization, and seems to permission to bypass the branch protection rules.

.github/workflows/cli-release.yml
```diff
@@ -0,0 +1,83 @@
...
+        run: yarn build
+
+      - name: Bump CLI version
+        run: npm version ${{ inputs.npm-version }} --force
+
+      - name: Get latest tag
+        run: echo "LATEST_TAG=$(git describe --abbrev=0 --tags)" >> $GITHUB_ENV
+          
+      - name: Commit and push tag to main branch
+        run: |
```
rob-odwyer, line 63: I might be wrong here, but I think you do need the email and name configured in git as well to push git changes - like you did already on your test fork.
I did this, but the way you tested with is probably just fine too:
```
git config --global user.email "github-tracker-bot@taplytics.com"
git config --global user.name "DevCycle Automation"
```

.github/workflows/cli-release.yml
```diff
@@ -0,0 +1,83 @@
...
+        required: false
+        type: choice
+        options: 
+          - major
+          - minor
+          - patch
+        default: 'patch'
+
+  # schedule:
+  #   - cron: '0 12 * * 2'  # Runs every Tuesday at 12:00 PM (noon) ET
```
rob-odwyer, line 27: Is this schedule something we're planning on doing? Or just an example?

.github/workflows/cli-release.yml
```diff
@@ -0,0 +1,83 @@
...
+      - name: Bump CLI version
+        run: npm version ${{ inputs.npm-version }} --force
+
+      - name: Get latest tag
+        run: echo "LATEST_TAG=$(git describe --abbrev=0 --tags)" >> $GITHUB_ENV
+          
+      - name: Commit and push tag to main branch
+        run: |
+          git add .
+          git commit --amend -m "Release $LATEST_TAG"
```
rob-odwyer, line 65: Do we want to use `--amend` here? I haven't used it myself, but my understanding was it it would replace the latest commit instead of making a new one.

rob-odwyer (commented):
Couple of small suggestions that I think might be needed for it to run successfully, but overall looks good!

.github/workflows/cli-release.yml
L-Joli: 😂👍🏻

.github/actions/release/action.yml
L-Joli: ohhh, I see, thanks for the update!

.github/workflows/cli-release.yml
L-Joli: I'm not sure too, since @JamieSinn commented earlier said the github user is auto configured🤔. I can add this first as I recognize it's from taplytics?🤔

.github/workflows/cli-release.yml
L-Joli: yes, I think the team is planning to do the release on schedule, but don't have a decision yet. Putting a comment here to remind myself and the team to bring this up again after this script is merged.

.github/workflows/cli-release.yml
L-Joli: yes, as the last commit created by npm version is not needed. I'm trying to combine the commit "Release $LATEST_TAG" and the npm version one together here

.github/workflows/cli-release.yml
JamieSinn: It may be needed after all - since we're bypassing and pushing to main

.github/workflows/cli-release.yml
rob-odwyer: Yeah, you might need it for both. I think it's needed anywhere you're pushing commits to a repo that has branch protection turned on.

.github/workflows/cli-release.yml
L-Joli: ok, thanks both! I'll make an update

rob-odwyer (approved):

JamieSinn (approved):

PR #225, opened 2023-07-18
fix: DVC-8206 update JS SDK references to new SDK repo names

jonathannorris:


src/commands/generate/types.test.ts
```diff
@@ -63,11 +63,11 @@ export type DVCVariableTypes = {
     'json-var': DVCJSON
 }`
 
-const expectedReactTypesString = `import { DVCVariable, DVCVariableValue } from '@devcycle/devcycle-js-sdk'
+const expectedReactTypesString = `import { DVCVariable, DVCVariableValue } from '@devcycle/js-client-sdk'
```
jonathannorris, line 66: The one question I had is should I update this script to try and figure out somehow which version of the JS SDKs they are using? Or are we ok with this just working going forward?

jsalaber (approved):

PR #229, opened 2023-07-19
chore: order env list by type in development, staging, production order

jsalaber:
# Changes

- order environment list when prompted to be in order of `development`, `staging`, `production` types.

### DIFF
--- a/src/ui/prompts/environmentPrompts.ts
+++ b/src/ui/prompts/environmentPrompts.ts
@@ -16,6 +16,13 @@ export type EnvironmentPromptResult = {
     environment: EnvironmentChoice['value']
 } & PromptResult
 
+const EnvironmentTypeValue: Record<string, number> = {
+    development: 0,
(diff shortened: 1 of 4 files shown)


src/ui/prompts/environmentPrompts.ts
```diff
@@ -16,6 +16,12 @@ export type EnvironmentPromptResult = {
     environment: EnvironmentChoice['value']
 } & PromptResult
 
+const EnvironmentTypeValue: Record<string, number> = {
```
elliotCamblor, line 19: how will this work if users have other envs/changed the keys?

laurawarr: this is the type rather than the key, but that reminds me, we also have the disaster_recovery type

elliotCamblor: oh true, good point

PR #234, opened 2023-07-19
Dvc 8235 headless environments

emmawillis:


src/commands/environments/get.ts
```diff
@@ -25,24 +25,30 @@ export default class DetailedEnvironments extends Base {
...
+
+        // prompt for keys in interactive mode 
+        const responses = await inquirer.prompt<EnvironmentPromptResult>(
+            [environmentPrompt],
+            {
+                token: this.authToken,
+                projectKey: this.projectKey
+            }
+        )
+        const environment = await fetchEnvironmentByKey(this.authToken, this.projectKey, responses.environment._id)
```
elliotCamblor, line 50: if you're fetching all of the environments anyway, does it make sense to just search those for this `_id` and save an api request?

emmawillis: good point

elliotCamblor (approved):

PR #235, opened 2023-07-19
chore: Add TokenCache tests

laurawarr:


src/auth/TokenCache.ts
```diff
@@ -27,7 +27,7 @@ export class TokenCache {
     public async get(clientId: string, clientSecret: string): Promise<string | null> {
```
elliotCamblor, line 27: not a big deal, but can the `async` be removed from the get / set methods now that you're using `read/writeFileSync`?

laurawarr: yep, good catch 👀

elliotCamblor (approved):

PR #237, opened 2023-07-28
feat: Use refresh token to fetch new token for sso auth

laurawarr:
- When fetching a new token with sso auth retrieve a refresh token as well
- When a token is within 1 hour of expiry and we have a refresh token available, fetch a new token

This allows users to login through the browser once, and then as long as they interact with the CLI within the refresh token expiry (15 days) the access token will be refreshed

### DIFF
--- a/oclif.manifest.json
+++ b/oclif.manifest.json
@@ -1,5 +1,5 @@
 {
-  "version": "4.6.3",
+  "version": "4.6.4",
   "commands": {
     "authCommand": {
       "id": "authCommand",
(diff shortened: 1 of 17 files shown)


oclif.manifest.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "4.6.3",
+  "version": "4.6.4",
```
jsalaber, line 2: is this cause the automated release doesn't update this?

laurawarr: Yeah, it must be. This was auto-updated by oclif on build

jsalaber (approved):

PR #245, opened 2023-08-02
Fetch keys flags individually instead of filtering

emmawillis:


src/commands/environments/get.ts
```diff
@@ -21,34 +21,34 @@ export default class DetailedEnvironments extends Base {
...
+        const keys = flags['keys']?.split(',')
         const { headless, project } = flags
         await this.requireProject(project, headless)
 
-        let environments = await fetchEnvironments(this.authToken, this.projectKey)
-        if (keys.length) {
-            environments = environments.filter((environment) => 
-                keys.includes(environment.key) || keys.includes(environment._id))
+        if (keys) {
+            const environments = await Promise.all(
```
laurawarr, line 29: Since a user could pass any number of keys, we should create batches of requests. You could checkout [lodash's chunk method](https://lodash.com/docs/4.17.15#chunk) for splitting the array into groups of requests

emmawillis: oh i see, would the idea be to prevent sending too many requests in parallel?

laurawarr: Yeah exactly, that way we can choose a reasonable number of requests to send in parallel (say ~10), and then even if the user passes 100+ keys we can await each batch rather than spamming our API

emmawillis: makes sense. i added that plus I added `.filter(Boolean)` to the results to avoid null/ undefined in the results for keys that aren't found

src/commands/environments/get.ts
```diff
@@ -21,34 +22,39 @@ export default class DetailedEnvironments extends Base {
...
-                keys.includes(environment.key) || keys.includes(environment._id))
+        if (keys) {
+            const environments = []
+            const batches = chunk(keys, 10)
+            for (const batch of batches) {
+                const batchEnvironments = await Promise.all(
+                    batch.map((key) => fetchEnvironmentByKey(this.authToken, this.projectKey, key))
+                )
+                environments.push(...batchEnvironments.filter(Boolean))
+            }
```
laurawarr, line 37: not necessary, but this could be a good place to create a util function that handles batching the requests given the array and map function

laurawarr (approved):
One suggestion, but otherwise looks good 👍

src/commands/environments/get.ts
emmawillis: good point, i moved it into utils

PR #247, opened 2023-08-03
fix: exclude context lines from git diff

maxnowack:
If there are more than 0 lines of context returned by `git diff', the CLI will also detect these context lines as changes, even though these lines have not been modified.
I noticed this bug while using the [feature-flag-pr-insights-action](https://github.com/DevCycleHQ/feature-flag-pr-insights-action). Some pull requests were commented by the action with variable changes (added and removed at the same time, but that is another issue), but the variables were not touched in the PR.

### DIFF
--- a/src/utils/diff/diff.ts
+++ b/src/utils/diff/diff.ts
@@ -4,7 +4,7 @@ import { readFileSync } from 'fs'
 
 export const executeDiff = (diffCommand: string): parse.File[] => {
     try {
-        execSync(`git diff ${diffCommand} > diff.txt`, { stdio: 'ignore' })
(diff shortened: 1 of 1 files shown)


laurawarr:
Hey @maxnowack, we rely on the context lines to help us find matches. For example, if a call to the variable method was split across multiple lines and only one was modified, we'd still consider that a change.
```
         this.dvcClient.variable(
             user,
-            'variable-key',
+            'new-variable',
             true
         )
```
For the case where the variable call wasn't changed at all, I'm testing with [this diff](https://github.com/DevCycleHQ/cli/blob/main/test-utils/fixtures/diff/no-change) and seeing the expected output, can you provide more details or a code snippet for the case you're describing?

PR #266, opened 2023-08-23
feat: add projects get command

nsalamad:


src/commands/projects/get.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export default class DetailedProjects extends Base {
+    static description = 'Retreive all projects in the current Organization'
+    static hidden = false
+    authRequired = true
+
+    public async run(): Promise<void> {
+        const projects = await fetchProjects(this.authToken)
+        return this.writer.showResults(projects.map((({ _id, key, name }) => (
+            { _id, key, name }
```
laurawarr, line 12: we could include other properties like description, createdAt, updatedAt etc.

jsalaber (approved):

README.md
```diff
@@ -100,7 +100,7 @@ $ npm install -g @devcycle/cli
 $ dvc COMMAND
 running command...
 $ dvc (--version)
-@devcycle/cli/5.5.0 linux-x64 node-v18.16.0
+@devcycle/cli/5.5.0 darwin-arm64 node-v18.16.0
```
laurawarr, line 103: shouldn't need to commit this change

src/commands/projects/get.ts
nsalamad: should we just include all the properties?

laurawarr (approved):
Added a couple small comments, otherwise looks good

src/commands/projects/get.ts
laurawarr: sure, if they're okay to return from the api, then they're fine to return from here 👍

PR #270, opened 2023-09-05
feat: add dvc identity get command

chantalchung:
- added `dvc identity get` command
- can take project key as well, otherwise gets identity for current project
cli-identity-get
    - chantal's project has a dvcUserId set, chantal 2.0 does not

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ USAGE
 * [`dvc features`](docs/features.md) - Create, view, or modify Features with the Management API.
 * [`dvc generate`](docs/generate.md) - Generate Devcycle related files.
 * [`dvc help`](docs/help.md) - Display help for dvc.
-* [`dvc identity`](docs/identity.md) - Update your DevCycle Identity.
+* [`dvc identity`](docs/identity.md) - View or update your DevCycle Identity.
 * [`dvc keys`](docs/keys.md) - Retrieve SDK keys from the Management API.
 * [`dvc login`](docs/login.md) - Log in to DevCycle.
 * [`dvc logout`](docs/logout.md) - Discards any auth configuration that has been stored in the auth configuration file.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -119,7 +119,7 @@ USAGE
 * [`dvc features`](docs/features.md) - Create, view, or modify Features with the Management API.
 * [`dvc generate`](docs/generate.md) - Generate Devcycle related files.
 * [`dvc help`](docs/help.md) - Display help for dvc.
-* [`dvc identity`](docs/identity.md) - Update your DevCycle Identity.
+* [`dvc identity`](docs/identity.md)
```
rabiaq123, line 122: was this supposed to have another description in place of the update one?

README.md
```diff
@@ -119,7 +119,7 @@ USAGE
 * [`dvc features`](docs/features.md) - Create, view, or modify Features with the Management API.
 * [`dvc generate`](docs/generate.md) - Generate Devcycle related files.
 * [`dvc help`](docs/help.md) - Display help for dvc.
-* [`dvc identity`](docs/identity.md) - Update your DevCycle Identity.
+* [`dvc identity`](docs/identity.md) - Print your DevCycle Identity.
```
rabiaq123, line 122: Maybe it should be "View or update" since `dvc identity update` is also an extension of the `dvc identity` command? Looked at the `dvc features` command for reference:)

docs/identity.md
```diff
@@ -1,10 +1,35 @@
 `dvc identity`
 ==============
 
-Update your DevCycle Identity.
```
rabiaq123, line 4: I'd put the "View or update" here as well

README.md
chantalchung: nice catch 👀 how's this for a description?
image

docs/identity.md
chantalchung: Updated 👀 for reference this part is defined in the package.json instead: https://github.com/DevCycleHQ/cli/pull/270/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R114-R116

kaushalkapasi (approved):

kaushalkapasi (approved):

PR #278, opened 2023-09-05
overrides get command

rabiaq123:
- `dvc overrides get --feature <feature-key> --environment <environment-key>` 
- use case: viewing a feature-specific override given the environment


src/api/zodClient.ts
```diff
@@ -3372,6 +3393,63 @@ const endpoints = makeApi([
...
         ],
     },
+    {
+        method: 'get',
+        path: '/v1/projects/:project/features/:feature/overrides/current',
+        alias: 'OverridesController_findOne',
+        requestFormat: 'json',
+        parameters: [
+            {
+                name: 'project',
```
kaushalkapasi, line 3403: missing feature param (also type `Path`) and environment param (type `Query`)

src/commands/overrides/get.ts
```diff
@@ -0,0 +1,91 @@
...
+        const overrides = await fetchFeatureOverridesForUser(this.authToken, this.projectKey, featureKey) 
+        const environment = await fetchEnvironmentByKey(this.authToken, this.projectKey, environmentKey)
+        const override = overrides.overrides.find((override) => override._environment === environment._id)
+
+        if (!override) {
+            if (headless) {
+                this.writer.showResults({ environment: environment.key, variation: null })
+                return
+            }
+            this.writer.showError(
```
kaushalkapasi, line 75: this should just be normal output, not `showError`

src/commands/overrides/get.ts
```diff
@@ -0,0 +1,92 @@
...
+} from '../../ui/prompts'
+import Base from '../base'
+import { fetchFeatureOverridesForUser } from '../../api/overrides'
+import { fetchEnvironmentByKey } from '../../api/environments'
+import { fetchVariationByKey } from '../../api/variations'
+
+export default class DetailedTargeting extends Base {
+    static hidden = false
+    authRequired = true
+    static description = 'View the overrides associated with your DevCycle Identity in your current project.'
```
chantalchung, line 17:
```suggestion
    static description = 'View the Overrides associated with your DevCycle Identity in your current project.'
```

src/commands/overrides/get.ts
```diff
@@ -0,0 +1,92 @@
...
+    EnvironmentPromptResult, 
+    featurePrompt, 
+    FeaturePromptResult 
+} from '../../ui/prompts'
+import Base from '../base'
+import { fetchFeatureOverridesForUser } from '../../api/overrides'
+import { fetchEnvironmentByKey } from '../../api/environments'
+import { fetchVariationByKey } from '../../api/variations'
+
+export default class DetailedTargeting extends Base {
```
chantalchung, line 14: should this be named something else?
```suggestion
export default class DetailedOverrides extends Base {
```

src/commands/overrides/get.ts
```diff
@@ -0,0 +1,92 @@
...
+
+        if (!override) {
+            if (headless) {
+                this.writer.showResults({ environment: environment.key, variation: null })
+                return
+            }
+            this.writer.showResults(
+                `Override for feature: ${featureKey} on environment: ${environment.key} is variation: <not-set>`
+            )
+            this.writer.infoMessageWithCommand('To set an override, use: ', 'dvc overrides update')
```
chantalchung, line 77: space is not needed here, it's handled in the function
```suggestion
            this.writer.infoMessageWithCommand('To set an override, use:', 'dvc overrides update')
```

src/commands/overrides/get.ts
```diff
@@ -0,0 +1,92 @@
...
+        )
+        const environment = await fetchEnvironmentByKey(this.authToken, this.projectKey, environmentKey)
+        const override = overrides.overrides.find((override) => override._environment === environment._id)
+
+        if (!override) {
+            if (headless) {
+                this.writer.showResults({ environment: environment.key, variation: null })
+                return
+            }
+            this.writer.showResults(
```
chantalchung, line 74: let's use `showRawResults` here so it doesn't stringify the result again and add extra quotations
image

suthar26: you mean line 74, just to be clear?

chantalchung: yup, in non-headless mode

PR #281, opened 2023-09-07
chore: remove variable feature prompt from create variable prompts, prompt manually in create command

jsalaber:
# Changes

- added prompt before prompting to attach to a feature to ask user if they want to attach variable to a feature

### DIFF
--- a/src/commands/variables/create.ts
+++ b/src/commands/variables/create.ts
@@ -3,9 +3,9 @@ import {
     createVariable,
 } from '../../api/variables'
 import CreateCommand from '../createCommand'
-import { createVariablePrompts } from '../../ui/prompts'
+import { createVariablePrompts, variableFeaturePrompt } from '../../ui/prompts'
 import { CreateVariableDto } from '../../api/schemas'
-import { ZodError } from 'zod'
+import inquirer from '../../ui/autocomplete'
 
 export default class CreateVariable extends CreateCommand {
     static hidden = false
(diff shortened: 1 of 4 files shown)


jsalaber:
* main
  * **PR #281** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/cli/281" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/cli/281?utm_source=stack-comment).

src/commands/variables/create.ts
```diff
@@ -38,7 +38,19 @@ export default class CreateVariable extends CreateCommand {
         }
         flags._feature = feature
 
-        const params = await this.populateParametersWithZod(CreateVariableDto, this.prompts, flags)
+        let params = await this.populateParametersWithZod(CreateVariableDto, this.prompts, flags)
+        if (!headless) {
+            const { attachToFeature } = await inquirer.prompt([{
+                name: 'attachToFeature',
+                message: 'Attach to a feature?',
```
ajwootto, line 45: is this the clearest way to phrase this? Should we use similar language to the dashboard like "associate"?

maybe explain what attaching a variable to a feature will do for you in case you didn't know?

like

`Would you like to associate this variable to a feature? Doing so will allow you to serve different values of the variable to different users"

jsalaber: good point, I think associate works better

elliotCamblor, line 45: Do you think it would make sense for `attach to` to instead be `associate with`? Only because we use the term "associated" in the dashboard/docs, otherwise I wouldnt pick this nit

jsalaber: yes I'll switch to associate

jsalaber: although giving a more detailed explanation would be weird to do it just for this one spot, and might also just be too wordy

elliotCamblor (approved):

PR #282, opened 2023-09-08
DVC-8696 [feat]: add ability to get all Overrides for a user in a project

kaushalkapasi:
- adds all flag to overrides GET command
- add table related flags to be used when getting all Project overrides
- adds ux.table output for all flags

image

### DIFF
--- a/src/api/zodClient.ts
+++ b/src/api/zodClient.ts
@@ -778,6 +778,15 @@ const FeatureOverride = z.object({
 const FeatureOverrideResponse = z.object({
     overrides: z.array(FeatureOverride)
 })
+const UserOverride = z.object({
+    _feature: z.string(),
+    featureName: z.string(),
+    _environment: z.string(),
+    environmentName: z.string(),
+    _variation: z.string(),
+    variationName: z.string(),
+})
+const UserOverrides = z.array(UserOverride)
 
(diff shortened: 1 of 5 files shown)


kaushalkapasi:
* main
  * **PR #282** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/cli/282" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/cli/282?utm_source=stack-comment).

suthar26 (approved):

src/api/zodClient.ts
```diff
@@ -778,6 +778,15 @@ const FeatureOverride = z.object({
 const FeatureOverrideResponse = z.object({
     overrides: z.array(FeatureOverride)
 })
+const UserOverride = z.object({
```
suthar26, line 781: do you not need to add this as a schema object? of is that only if we need to use the type elsewhere?

kaushalkapasi: added to the object in case we need it in the future, but not needed since this is response type and not a payload type / something that we'd use to validate sending payloads to the API

suthar26: that makes sense!

PR #283, opened 2023-09-11
DVC-8773 [feat]: add overrides list command

kaushalkapasi:
- feat: adds overrides list command
- chore: fix output for overrides get table output
- chore: fix output for tables to include newline after for better readability

### DIFF
--- a/src/commands/overrides/get.ts
+++ b/src/commands/overrides/get.ts
@@ -7,9 +7,10 @@ import {
     FeaturePromptResult
 } from '../../ui/prompts'
 import Base from '../base'
-import { fetchFeatureOverridesForUser, fetchProjectOverridesForUser } from '../../api/overrides'
+import { fetchFeatureOverridesForUser } from '../../api/overrides'
 import { fetchEnvironmentByKey } from '../../api/environments'
 import { fetchVariationByKey } from '../../api/variations'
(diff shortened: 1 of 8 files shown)


src/commands/overrides/get.ts
```diff
@@ -107,8 +88,13 @@ export default class DetailedOverrides extends Base {
             return
         }
 
-        this.writer.successMessage(
-            `Override for feature: ${featureKey} on environment: ${environment.key} is variation: ${variation.key}`
-        )
+        this.tableOutput.printOverrides<UserOverride>([{
```
suthar26, line 91: why are we not passing in the options with this?

src/commands/overrides/get.ts
```diff
@@ -35,31 +31,15 @@ export default class DetailedOverrides extends Base {
             description: 'The key or id of the Environment to get Overrides for',
         }),
         ...Base.flags,
-        ...ux.table.flags(),
```
suthar26, line 38: any reason for removing these?

src/commands/overrides/get.ts
kaushalkapasi: its a single entry, so doesn't make sense to pass in the table flags to sort / filter since there's only one item being rendered at most

suthar26 (approved):

PR #284, opened 2023-09-11
DVC-8772 chore: filter prompts for overrides

suthar26:
- tested locally by changing the number of overrides through dashboard 

Screenshot 2023-09-11 at 2 40 44 PM

### DIFF
--- a/src/ui/prompts/overridePrompts.ts
+++ b/src/ui/prompts/overridePrompts.ts
@@ -1,0 +1,72 @@
+import { fetchEnvironments } from '../../api/environments'
+import { fetchFeatures } from '../../api/features'
+import { fetchFeatureOverridesForUser, fetchProjectOverridesForUser } from '../../api/overrides'
+import { Environment, Feature } from '../../api/schemas'
+import { autocompleteSearch } from '../autocomplete'
+import { EnvironmentChoice } from './environmentPrompts'
+import { FeatureChoice } from './featurePrompts'
+
+type Input = {
+    token: string,
+    projectKey: string,
+    featureKey?: string,
+}
+export const overridesFeatureChoices =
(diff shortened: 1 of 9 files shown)


suthar26:
* main
  * **PR #284** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/cli/284" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/cli/284?utm_source=stack-comment).

src/ui/prompts/overridePrompts.ts
```diff
@@ -0,0 +1,72 @@
...
+            }
+            if (choices.find((choice: FeatureChoice) => choice.value._id === feature._id)) {
+                continue
+            }
+            choices.push({
+                name: feature.name || feature.key,
+                value: feature
+            })
+        }
+        return autocompleteSearch(choices, search)
```
chantalchung, line 33: I think we need some kind of empty state message if there are no features with overrides (`choices` array is empty) - same case for environments below
image
I'm stuck on this step when I have no overrides

suthar26: yess will add!

suthar26: this is the same thing that happens when you create a project with no features and do `feature update` the list is empty.. couldn't find a way to make it not not execute here without making an extra call to the api

kaushalkapasi: just escape early, if no results found then return and say `No Overrides found.` with an info print

suthar26: added empty states

PR #285, opened 2023-09-11
chore: overrides update command to print results in table

chantalchung:
- updated the `overrides update` command to print the results in a table
image

### DIFF
--- a/src/commands/overrides/update.ts
+++ b/src/commands/overrides/update.ts
@@ -10,6 +10,7 @@ import { UpdateOverrideDto } from '../../api/schemas'
 import { updateOverride } from '../../api/overrides'
 import Base from '../base'
 import { fetchUserProfile } from '../../api/userProfile'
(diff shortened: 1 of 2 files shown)


src/commands/overrides/update.ts
```diff
@@ -80,16 +86,20 @@ export default class UpdateOverride extends Base {
...
             feature, environment, variation
         })
 
         const result = await updateOverride(this.authToken, this.projectKey, feature, params)
-        this.writer.successMessage(
-            `Successfully added Override for User ID: ${result.dvcUserId}` +
-            ` for feature: ${feature} on environment: ${environment}` +
-            ` to variation: ${variation}`)
+        const resultToPrint = { 
+            featureName: featureName ?? '',
```
kaushalkapasi, line 98: fallback to the `key` for all 3?

kaushalkapasi (approved):

PR #291, opened 2023-09-14
feat: add info comments to type generator

nsalamad:
image

### DIFF
--- a/src/commands/generate/types.ts
+++ b/src/commands/generate/types.ts
@@ -3,6 +3,7 @@ import fs from 'fs'
 import { fetchAllVariables } from '../../api/variables'
 import { Flags } from '@oclif/core'
 import { Variable } from '../../api/schemas'
+import { OrganizationMember, fetchOrganizationMembers } from '../../api/members'
 
 const reactImports = (oldRepos: boolean) => {
     const jsRepo = oldRepos ? '@devcycle/devcycle-js-sdk' : '@devcycle/js-client-sdk'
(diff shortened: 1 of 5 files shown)


src/commands/generate/types.ts
```diff
@@ -110,4 +142,45 @@ export default class GenerateTypes extends Base {
         }
         return variable.type.toLocaleLowerCase()
     }
+
+    private async getVariableInfoComment(variable: Variable, orgMembers: OrganizationMember[]): Promise<string> {
+        const flags = await this.parse(GenerateTypes)
+        const { 'include-descriptions': includeDescriptions, 'inline-comments': inlineComments } = flags.flags
```
elliotCamblor, line 148: you're accessing `flags.flags` here instead of just doing `const { flags }` in the line above (I only mention this because this is how you're accessing flags in the `getTypeDefinitionLine function` so its inconsistent)

src/commands/generate/types.ts
```diff
@@ -80,18 +90,40 @@ export default class GenerateTypes extends Base {
...
+            typeLines.join('\n') +
             '\n}'
         if (react) {
             types += reactOverrides
         }
         return types
     }
 
+    private async getTypeDefinitionLine(variable: Variable, orgMembers: OrganizationMember[]): Promise<string> {
+        const { flags } = await this.parse(GenerateTypes)
```
elliotCamblor, line 113: Does it make sense to just pass the flags around this these functions rather than having to parse them out 2-3 times?

elliotCamblor (commented):
Left some (mostly) nitpicky comments

src/commands/generate/types.ts
nsalamad: I personally don't like the idea of passing the args around cause it clutters the definitions/calls and makes it more likely that someone will mistakenly pass the args in the wrong positions. Like some of the methods are gonna end up having lots of args i.e. `getTypesString` with 6 which is too much imo

nsalamad: was even considering changing `getTypesString` to parse as well

PR #297, opened 2023-09-26
feat: add command to display current project

nsalamad:
`projects current`
image

`status`
image

`projects list`
image

`keys get`
image

`login again`
image

### DIFF
--- a/src/commands/base.ts
+++ b/src/commands/base.ts
@@ -4,7 +4,7 @@ import { Command, Flags } from '@oclif/core'
 import fs from 'fs'
 import path from 'path'
 import jsYaml from 'js-yaml'
-import { RepoConfigFromFile, UserConfigFromFile } from '../types'
+import { RepoConfigFromFile, SavedOrganization, UserConfigFromFile } from '../types'
 import { ClassConstructor, plainToClass } from 'class-transformer'
 import { validateSync } from 'class-validator'
 import { reportValidationErrors, reportZodValidationErrors, validateParams } from '../utils/reportValidationErrors'
@@ -78,6 +78,7 @@ export default abstract class Base extends Command {
     personalAccessToken = ''
     projectKey = ''
     orgId = ''
+    organization: SavedOrganization | undefined = undefined 
     authPath = path.join(this.config.configDir, 'auth.yml')
     configPath = path.join(this.config.configDir, 'user.yml')
     repoConfigPath = '.devcycle/config.yml'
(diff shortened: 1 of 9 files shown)


src/commands/keys/get.ts
```diff
@@ -59,6 +60,12 @@ export default class GetEnvironmentKey extends Base {
         if (flags.env) {
             return flags.env
         }
+        const organization = this.repoConfig?.org || this.userConfig?.org
+        const project = this.repoConfig?.project || this.userConfig?.project
```
laurawarr, line 64: Ideally we could display project/org property defined on the Base command. This command shouldn't need to know where the project/org came from (ie. repo vs user config), and it should always match what the base command is using, even if that logic changes in the future

src/commands/status/index.ts
```diff
@@ -48,7 +50,12 @@ export default class ShowStatus extends Base {
         const loggedInOrg = this.repoConfig?.org || this.userConfig?.org
         if (this.hasToken()) {
             if (loggedInOrg) {
-                this.writer.successMessage(`Currently logged in to DevCycle as a member of ${loggedInOrg.display_name}`)
+                this.writer.successMessage('Logged into DevCycle with:')
+                this.writer.showRawResults(chalk.green(`\tOrganization: ${loggedInOrg.display_name}`))
+                const loggedInProject = this.repoConfig?.project || this.userConfig?.project
```
laurawarr, line 55: Same thing here — it seems like we were already duplicating the repo/user config logic in a few places, but if you look at the Base command the project and org can be defined in more ways than just the configs. We should rely on these properties to tell us which project/org is the current one
```
this.projectKey = flags['project']
            || process.env.DEVCYCLE_PROJECT_KEY
            || process.env.DVC_PROJECT_KEY
            || this.repoConfig?.project
            || this.userConfig?.project
            || ''
this.orgId = flags['org'] || this.repoConfig?.org?.id || this.userConfig?.org?.id
```

laurawarr:
Nice work! I like that you included the project on all these commands rather than just the status 👌

src/commands/status/index.ts
nsalamad: looks like there's no `org` flag on the base class so we should drop the `flags['org']` in
```
this.orgId = flags['org'] || this.repoConfig?.org?.id || this.userConfig?.org?.id
```
Also I don't think letting users pass an org as flag makes sense since trying to use an org id other than the one that you have authorized with would result in auth errors.

Let me know if you think otherwise

laurawarr: The org flag is used by any command that extends the AuthCommand, it will use the specified org for authentication rather than org in the config (ie. `login again`, `login sso`, `organizations select`, `projects select`, `repo init`)

src/commands/base.ts
```diff
@@ -242,6 +243,7 @@ export default abstract class Base extends Command {
             || this.userConfig?.project
             || ''
         this.orgId = flags['org'] || this.repoConfig?.org?.id || this.userConfig?.org?.id
+        this.organization_name = this.repoConfig?.org?.display_name || this.userConfig?.org?.display_name || ''
```
laurawarr, line 246: you could make this a little more generic, so that we can access any properties
```suggestion
        this.organization = this.repoConfig?.org || this.userConfig?.org
        this.orgId = flags['org'] || this.organization?.id
```

laurawarr (approved):
added one comment, otherwise looks good 👍

PR #303, opened 2023-09-28
chore: Update min node version to 16.0.0

laurawarr:
oclif uses the AbortController when parsing flag values which was added in node v15

### DIFF
--- a/package.json
+++ b/package.json
@@ -133,7 +133,7 @@
     "build:watch": "watch 'yarn build' src"
   },
   "engines": {
-    "node": ">=14.18.0"
+    "node": ">=16.0.0"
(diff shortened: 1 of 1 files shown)


kaushalkapasi (approved):

package.json
```diff
@@ -133,7 +133,7 @@
     "build:watch": "watch 'yarn build' src"
   },
   "engines": {
-    "node": ">=14.18.0"
+    "node": ">=16.0.0"
```
kaushalkapasi, line 136: Note: Update here to node `>=16.0.0` instead of just Node v15 due to LTS support of node versions

PR #304, opened 2023-09-29
chore: show variation name when updating feature variables

nsalamad:


src/ui/prompts/variablePrompts.ts
```diff
@@ -132,13 +131,13 @@ export const getVariableValuePrompt = (
 ) => {
     switch (variableType) {
         case 'Boolean':
-            return variableValueBooleanPrompt(variation.key, defaultValue as boolean | undefined)
+            return variableValueBooleanPrompt(variation.key, defaultValue as boolean | undefined, variation.name)
```
laurawarr, line 134: might be more clear to just pass the full variation as a param?

nsalamad: since these methods are also used by `variations update` and the cleanup command, i had to use discriminated unions to differentiate when to access `name`. It's a little more clean than passing separate params

src/ui/prompts/variablePrompts.ts
```diff
@@ -50,20 +55,30 @@ export const variableTypePrompt = {
...
+): Prompt => {
+    const key = entity.value.key
+    const messageIdentifier = entity.type === 'variation' ? entity.value.name : entity.value.key
     return {
-        name: variableKey,
+        name: key,
         default: defaultValue,
         type: 'input',
-        message: `Variable value for ${variableKey}`,
+        message: `Variable value for ${messageIdentifier}`,
```
laurawarr, line 68: it seems like this is duplicated for each type of prompt? we could create a utility method for building the base prompt and then add any additional properties specific to each type

nsalamad: yup agreed

src/ui/prompts/variablePrompts.ts
```diff
@@ -132,13 +157,25 @@ export const getVariableValuePrompt = (
 ) => {
     switch (variableType) {
         case 'Boolean':
-            return variableValueBooleanPrompt(variation.key, defaultValue as boolean | undefined)
+            return variableValueBooleanPrompt(
```
laurawarr, line 160: not related to your change, but we could overload the `getVariableValuePrompt` method to make sure the variable type and default value match and then remove the type casting on each of these calls

laurawarr (approved):

src/ui/prompts/variablePrompts.ts
nsalamad: good idea! I'll make a low prio ticket for it

PR #309, opened 2023-10-02
Dvc 8568 variations update in variables create

elliotCamblor:
variables_create


README.md
```diff
@@ -100,7 +100,7 @@ $ npm install -g @devcycle/cli
 $ dvc COMMAND
 running command...
 $ dvc (--version)
-@devcycle/cli/5.11.1 linux-x64 node-v18.16.0
```
emmawillis, line 103: are we supposed to be committing these changes? my env always generates them but i just delete them

src/commands/variables/create.ts
```diff
@@ -29,29 +35,73 @@ export default class CreateVariable extends CreateCommand {
...
+                feature.variables.push(params)
+                const variableListOptions = new VariableListOptions(
+                    [params as Variable],
+                    this.writer,
+                    !variations ? feature.variations: undefined // don't prompt for variations if flag provided
+                )
+                await variableListOptions.promptVariationValues(params as Variable)
+                await updateFeature(this.authToken, this.projectKey, feature.key, feature)
+                const message = `The variable was associated to the existing feature ${feature.key}. ` +
+                    `Use "dvc feature get ${feature.key}" to see its details`
```
emmawillis, line 78: should be `dvc features get --keys=${feature.key}`, i don't think we support `get <key>` currently (although we should)

emmawillis: also should we still output the variable result here as well as the success message? might be kind of awkward since we'd have to find the new variable in the result of `updateFeature` but i feel like it might be annoying to have to get the feature separately just to check if you made a typo or something

elliotCamblor: tbh I dont really think outputting the full variable is that useful here. Adam suggested that this be the response and I think I agree with him. Checking for a type is pretty much the only use case (especially since the command doesnt support headless mode if youre creating a feature)

elliotCamblor: I will change the command thats output though, good catch!

emmawillis: makes sense, works for me!

emmawillis (approved):

PR #310, opened 2023-10-02
feat: add sortBy and sortOrder flags to projects get command

L-Joli:
- add `sortBy` and `sortOrder` flags to projects get command

### DIFF
--- a/src/commands/getCommand.ts
+++ b/src/commands/getCommand.ts
@@ -1,0 +1,26 @@
+import Base from './base'
+import { Flags } from '@oclif/core'
+
+export default abstract class GetCommand extends Base {
+    authRequired = true
+
+    static flags = {
+        ...Base.flags,
+        'sortBy': Flags.string({
+            description: 'Sort By',
+            options: [
+                'key',
+                'name',
+                'updatedAt',
+                'createdAt'
+            ],
+        }),
+        'sortOrder': Flags.string({
+            description: 'Sort Order',
+            options: [
+                'asc',
+                'desc',
+            ],
+        }),
+    }
+}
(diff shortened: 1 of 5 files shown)


src/commands/getCommand.ts
```diff
@@ -0,0 +1,27 @@
+import Base from './base'
+import { Flags } from '@oclif/core'
+
+export default abstract class GetCommand extends Base {
+    authRequired = true
+
+    static flags = {
+        ...Base.flags,
+        'sortBy': Flags.string({
```
jsalaber, line 9: should we just sort by name by default and not need to specify in the command? since this will likely need a breaking change if it changes the behaviour of the default gets if calling `get` will prompt the user

src/commands/getCommand.ts
```diff
@@ -0,0 +1,27 @@
...
+            description: 'Sort By',
+            options: [
+                '_id',
```
nsalamad, line 12: I'm not sure sorting by mongo id would be very useful. We can probably get rid of this as an option

L-Joli: yes, I'm not sure about that too, but since our api accepts this option, do you think we should keep it here too?

nsalamad: technically our api accepts all the fields as options for sorting. But i  think we should remove this as an option. It's an internal db field for the project so i don't think sorting by this would useful

src/commands/projects/get.ts
```diff
@@ -1,13 +1,30 @@
...
+    static flags = {
+        ...GetCommand.flags,
+    }
+
     public async run(): Promise<void> {
-        const projects = await fetchProjects(this.authToken)
+        const { flags } = await this.parse(DetailedProjects)
+        const { sortBy, sortOrder, headless } = flags
+
+        const params = await this.populateParameters(
```
nsalamad, line 17: Is calling `populateParameters `still necessary if we're no longer prompting the user? can't the sorting flags be passed directly to `fetchProjects` if they're defined?

nsalamad (approved):
👍

PR #353, opened 2024-03-05
Add Prettier

ajwootto:


README.md
```diff
@@ -10,31 +9,39 @@ The CLI can be customized in several ways using command-line args or by creating
 [![License](https://img.shields.io/npm/l/@devcycle/cli.svg)](https://github.com/DevCycleHQ/cli/blob/main/package.json)
 
 <!-- toc -->
+* [DevCycle CLI](#devcycle-cli)
```
suthar26, line 12: this doesnt seem like it should be here

test-utils/fixtures/cleanup/javascript/[org_id]/test.js
Comment (suthar26) [inline test-utils/fixtures/cleanup/javascript/[org_id]/test.js:1]: should we have both single quotes and double quotes in the test utils files?

Comment (ajwootto) [reply test-utils/fixtures/cleanup/javascript/[org_id]/test.js:1]: not sure what you mean, there's only single quotes here

Comment (suthar26) [reply test-utils/fixtures/cleanup/javascript/[org_id]/test.js:1]: the strings were double quotes and now changes to single quotes, should we keep use of both and remove these cleanup files from prettier? would help run tests between both usecases

Comment (ajwootto) [reply test-utils/fixtures/cleanup/javascript/[org_id]/test.js:1]: going to ignore the fixtures directory from prettier

suthar26 (approved):

PR #354, opened 2024-03-05
obfuscation

ajwootto:
Add `--obfuscate` flag to `generate types` command allowing you to generate a set of obfuscated strings


ajwootto:
* **#354** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/cli/354?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @ajwootto and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

src/commands/generate/types.ts
```diff
@@ -69,24 +68,55 @@ export default class GenerateTypes extends Base {
...
+        this.includeDescriptions = includeDescriptions
+        this.inlineComments = inlineComments
+        this.obfuscate = obfuscate
+        this.project = await this.requireProject(project, headless)
+
+        if (
+            this.project.settings.obfuscation.enabled &&
+            this.project.settings.obfuscation.required
+        ) {
+            this.obfuscate = true
```
laurawarr, line 104: if `obfuscate` was explicitly passed as `false` it might be nice to log a warning that it's required and will be set to `true` regardless of the flag value

src/commands/base.ts
```diff
@@ -304,6 +315,14 @@ export default abstract class Base extends Command {
                 await this.updateUserConfig({ project: this.projectKey })
             }
         }
+
+        const project = projects.find((proj) => proj.key === this.projectKey)
```
jsalaber, line 319: can maybe make finding a project a util function

src/api/zodClient.ts
```diff
@@ -23,10 +23,16 @@ const OptInSettings = z.object({
 const SDKTypeVisibilitySettings = z.object({
     enabledInFeatureSettings: z.boolean(),
 })
+const ObfuscationSettings = z.object({
+    enabled: z.boolean(),
+    required: z.boolean(),
+    key: z.array(z.number()),
```
laurawarr, line 29: What is the `key` property used for? This doesn't seem to be added in your change to the API entity

ajwootto: need to remove this

src/commands/base.ts
ajwootto: to me it seems like too simple of a call to make it a separate func

.gitignore
```diff
@@ -7,3 +7,4 @@ dist
 .prettierrc
 tmp
 .yarn/cache
+.yarn/install-state.gz
```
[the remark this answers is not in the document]
ajwootto: its removed by the PR

src/commands/base.ts
jsalaber: can combine it with the below call that throws, or just make the error message a constant so it's not repeated

ajwootto: updated lmk what you think

jsalaber (approved):

PR #363, opened 2024-03-13
add option to automatically detect aliases from type files

ajwootto:
When generating the types file with the CLI, it now also generates a set of constants you can use in variable calls instead of hard-coding the strings.

This now records where that type file was generated to in the repo config, and loads it during `usages` detection to automatically detect aliases that should be used.


src/commands/generate/types.ts
```diff
@@ -230,6 +242,24 @@ export const ${constantName} = '${hashedKey}' as const`
     private encryptKey(variable: Variable) {
         return `dvc_obfs_${createHash('sha256').update(variable._id).digest('hex')}`
     }
+
+    private updateAliasesInRepo() {
```
laurawarr, line 246: shouldn't this method only be called if `updateAliases` is true?

ajwootto: ya

src/commands/generate/types.ts
```diff
@@ -230,6 +244,24 @@ export const ${constantName} = '${hashedKey}' as const`
...
+        if (!this.repoConfig) {
+            if (this.updateAliases) {
+                this.writer.failureMessage(
+                    'Repo configuration not found, aliases will not be updated',
+                )
+            }
+            return
+        }
```
laurawarr, line 256: this could be simplified to write the error any time the repo config doesn't exist, since we're assuming `updateAliases` is true

src/flags/var-alias/index.ts
```diff
@@ -13,8 +14,51 @@ export function getVariableAliases(
     flags: Record<string, any>,
     config: RepoConfigFromFile | null,
 ) {
-    const variableAliasesFromConfig =
-        config?.codeInsights?.variableAliases || {}
+    const variableAliases: Record<string, string> =
+        config?.codeInsights?.variableAliases ?? {}
+    const typesFileLocation = config?.typeGenerator?.fileLocation
+
+    if (typesFileLocation) {
```
laurawarr, line 21: it seems like this block of logic should be pulled out into a util function to avoid cluttering this method

src/commands/generate/types.ts
```diff
@@ -230,6 +239,34 @@ export const ${constantName} = '${hashedKey}' as const`
...
+            })
+            if (fileLocation) {
+                this.writer.successMessage(
+                    `Updated configured types file location to ${newFileLocation}`,
+                )
+            } else {
+                this.writer.successMessage(
+                    `Stored configured types file location as ${newFileLocation}`,
+                )
+            }
```
laurawarr, line 267: Just a nitpick, but it seemed unnecessarily verbose
```suggestion
            const message = fileLocation
                ? `Updated configured types file location to ${newFileLocation}`
                : `Stored configured types file location as ${newFileLocation}`
            this.writer.successMessage(message)
```

src/flags/var-alias/index.ts
```diff
@@ -26,6 +32,46 @@ export function getVariableAliases(
             }
             return { ...map, [codeRef]: variableName }
         },
-        variableAliasesFromConfig,
+        variableAliases,
     )
 }
+
+const addAliasesFromTypeGenerator = (
```
laurawarr, line 39: it'd probably be a good idea to add some unit tests to verify this continues to work as expected

ajwootto: i couldnt find any existing tests for the aliases system, do you know if there are any

laurawarr: I think most of the alias tests are within the command-specific tests. In this case I could see value in a colocated unit test that validates this method specifically

ajwootto: added some tests if you want to take another look

laurawarr (approved):

PR #390, opened 2024-05-09
fix: pagination bug fix

suthar26:
- tested with perPage as 5 to test locally and ensure the type generation works as expected

### DIFF
--- a/src/api/members.ts
+++ b/src/api/members.ts
@@ -7,32 +7,35 @@ export type OrganizationMember = {
     email?: string | undefined
 }
 
-export const fetchOrganizationMembers = async (token: string): Promise<OrganizationMember[]> => {
+export const fetchOrganizationMembers = async (
+    token: string,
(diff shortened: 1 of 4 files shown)


src/api/members.ts
```diff
@@ -7,32 +7,36 @@ export type OrganizationMember = {
...
+    const promises = []
+    const totalPages = Math.ceil(total / perPage)
+    for (let i = 2; i <= totalPages; i++) {
+        promises.push(fetchMembers(i))
+    }
+    await Promise.all(promises).then((responses) => {
+        responses.forEach((response) => {
+            members.push(...response.data)
+        })
+    })
```
kaushalkapasi, line 39: Add a `.catch(err)` and handle any errors

suthar26: changed away from .then()

kaushalkapasi (approved):

PR #400, opened 2024-06-06
fix: renamed env variable to prevent collisions

jsalaber:
## Summary by Sourcery

This pull request addresses a potential environment variable collision by renaming 'LOCAL' to 'DVC_API_LOCAL' in the API configuration.

* **Bug Fixes**:
    - Renamed environment variable from 'LOCAL' to 'DVC_API_LOCAL' to prevent collisions.

### DIFF
--- a/src/api/common.ts
+++ b/src/api/common.ts
@@ -1,5 +1,5 @@
 export const BASE_URL =
-    process.env.LOCAL === '1'
+    process.env.DVC_API_LOCAL === '1'
         ? 'http://localhost:4001'
         : 'https://api.devcycle.com'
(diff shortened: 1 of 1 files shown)


kaushalkapasi (approved):

src/api/common.ts
```diff
@@ -1,5 +1,5 @@
 export const BASE_URL =
-    process.env.LOCAL === '1'
+    process.env.DVC_API_LOCAL === '1'
```
kaushalkapasi, line 2: do we need a readme update as well?

jsalaber: it is not in the readme

PR #401, opened 2024-06-11
fix: added retry for connecting to a port not in use

jsalaber:
# Changes

- when port is in use, add retry with another supported port to connect to

### DIFF
--- a/src/auth/SSOAuth.ts
+++ b/src/auth/SSOAuth.ts
@@ -18,8 +18,8 @@ type OauthResponse = {
         refresh_token: string
     }
 }
-const SUPPORTED_PORTS = [80, 8080, 2194, 2195, 2196]
-const PORT = Number.parseInt(process.env.PORT || '2194')
+let SUPPORTED_PORTS = [8080, 2194, 2195, 2196]
+let PORT = Number.parseInt(process.env.PORT || '2194')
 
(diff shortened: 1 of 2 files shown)


suthar26 (approved):

kaushalkapasi (approved):

src/auth/SSOAuth.ts
```diff
@@ -119,10 +119,27 @@ export default class SSOAuth {
...
+                if (!SUPPORTED_PORTS.length) {
+                    this.writer.showError(
+                        'No available ports to start local server',
+                    )
+                    exit(1)
+                }
+                PORT = SUPPORTED_PORTS[0]
+                setTimeout(() => {
+                    this.server.close()
+                    this.server.listen(PORT, host, () => {
```
ajwootto, line 132: are you sure there isn't a way to just tell it to listen on "any available open port"? Usually if you just pass 0 or something then it will do that

jsalaber: our auth is setup to only accept the supported ports as redirects

jsalaber: so using any port will unfortunately error

ajwootto: gotcha

PR #409, opened 2024-07-01
fix: Move `mocha-chai-jest-snapshot` to be a dev dependency

csandman:
This is a small fix for an issue I've been having for a while installing this package as a dev dependency of my team's app. Every time I do a `yarn install` in my project, I get a warning about a missing peer dependency of one of this package's dependencies, `mocha-chai-jest-snapshot`. The fact that this is even considered a dependency and not a dev dependency doesn't really make sense, as it's just a testing library, and all packages that only have to do with testing should not be in the normal dependencies.

So this PR just moves it from being a dependency to a dev dependency. I thought about making an issue for this, but the fix was so straightforward, I figured I'd jus put up a PR instead. Let me know if this works for you!

Here's the actual warning by the way:

Screenshot 2024-07-01 at 2 14 05 PM

```
warning "@devcycle/cli > mocha-chai-jest-snapshot@1.1.4" has unmet peer dependency "chai@^4.3.6".
```


oclif.manifest.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "5.14.12",
+  "version": "5.14.13",
```
csandman, line 2: This was updated when I ran the build command. Not sure what `oclif` is used for tbh, but when I ran the tests after making my change, one of them failed because of this version, so I included this in the PR.

PR #410, opened 2024-07-02
Remove dependency on js-client-sdk for React generated types

emmawillis:


src/commands/generate/__snapshots__/types.test.ts.snap
```diff
@@ -185,10 +185,11 @@ export const JSON_VAR = 'dvc_obfs_a991871dbce45b30da1e125ae28b59eb74c1da1f48ad6a
 `;
 
 exports[`generate types correctly generates React SDK types 1`] = `
-"import { DVCVariable, DVCVariableValue } from '@devcycle/js-client-sdk'
-import {
+"import {
     useVariable as originalUseVariable,
-    useVariableValue as originalUseVariableValue
+    useVariableValue as originalUseVariableValue,
+    DVCVariable,
```
ajwootto, line 191: do these have "DevCycle" equivalents? We should be importing DevCycleVariable if that exists

emmawillis: i thought we had them too, but looks like the types are all still DVC___ currently 
https://github.com/DevCycleHQ/js-sdks/blob/main/sdk/js/src/types.ts#L198

ajwootto (approved):

PR #424, opened 2024-08-26
update generator to use new DevCycleJSON type

ajwootto:


kaushalkapasi (approved):

src/commands/generate/types.ts
```diff
@@ -383,7 +391,7 @@ export function getVariableType(variable: Variable) {
         return enumValues.map((value) => `'${value}'`).join(' | ')
     }
     if (variable.type === 'JSON') {
-        return 'DVCJSON'
+        return 'DevCycleJSON'
```
kaushalkapasi, line 394: Any concern with this resulting in misleading error for older installations of the cli?

ajwootto: not sure what you mean, this won't be in the old versions of the CLI. if you mean older SDKs then yeah it'll cause an error until the SDK is updated

kaushalkapasi: yeah, older SDK with new CLI.. but yeah, it would be odd to update one but not the other 🙃

PR #427, opened 2024-08-28
fix: update buffer size to 10MB

suthar26:
update the size so the larger repos dont fail while running usages
tested on a repo with ~50000 files


elliotCamblor (approved):

ajwootto:
doesn't this just make it stop scanning files in the repo after a certain point? like what happens to the rest of the file names that dont fit in the buffer?

PR #428, opened 2024-08-30
chore: add error message if ls-files command fails

suthar26:


src/utils/git/ls-files.ts
```diff
@@ -1,13 +1,20 @@
...
+        const fileBuffer = execSync(
+            'git ls-files --others --cached --exclude-standard ',
+            { maxBuffer: 1024 * 1024 * 10 }, // 10 MB buffer
+        )
+        const files = fileBuffer.toString().split('\n').filter(Boolean)
 
-    return files
+        return files
+    } catch (error) {
+        console.error(
```
ajwootto, line 15: should it check if the error is ENOBUF?

suthar26: yeah added the check for this

ajwootto (approved):

PR #430, opened 2024-09-03
add deprecation notices to type generator output when feature is complete

ajwootto:
- check if each variable is part of a completed or archived feature and is not marked permanent
- if all of that is true, add a deprecation notice to the variable and its constant alias
- remove option for inline comments because it's not really that useful and requires maintaining two different code paths for comment generation, plus deprecation warnings dont work in that format
image

### DIFF
--- a/src/api/features.ts
+++ b/src/api/features.ts
@@ -1,5 +1,5 @@
 import { AxiosError } from 'axios'
-import apiClient from './apiClient'
+import apiClient, { axiosClient } from './apiClient'
 import { CreateFeatureParams, Feature, UpdateFeatureParams } from './schemas'
 import 'reflect-metadata'
 import { buildHeaders } from './common'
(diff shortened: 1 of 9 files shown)


src/api/features.ts
```diff
@@ -91,3 +91,44 @@ export const deleteFeature = async (
...
+    return allFeatures.flat()
+}
+
+const generatePaginatedFeatureUrl = (
+    project_id: string,
+    page: number,
+    perPage: number,
+    status: string
+): string => {
+    return `/v1/projects/${project_id}/features?perPage=${perPage}&page=${page}&status=${status}`
```
jonathannorris, line 133: any way for us to page based on last `_id`?

ajwootto: no the API does not currently support that

src/commands/generate/types.ts
```diff
@@ -395,3 +376,9 @@ export function getVariableType(variable: Variable) {
     }
     return variable.type.toLocaleLowerCase()
 }
+
+function isVariableDeprecated(variable: Variable, features: Feature[]) {
+    if (!variable._feature || variable.persistent) return { deprecated: false}
+    const feature = features.find(f => f._id === variable._feature)
+    return {deprecated: feature && feature.status !== 'active', feature}
```
jonathannorris, line 383: wouldn't this catch `in progress` features?

ajwootto: there are only three statuses, active, complete and archived. It's also only fetching completed and archived features to begin with

jonathannorris: kk, was just looking at the ones we list on the dashboard: 
image

jonathannorris (approved):

PR #435, opened 2024-12-03
update CLI for declaration merging

ajwootto:
- update CLI to generate declare module interface merge that takes advantage of new type definitions
- add an option to disable the interface output
- deprecate the `--react` and `--next` flags

### DIFF
--- a/src/commands/generate/__snapshots__/types.test.ts.snap
+++ b/src/commands/generate/__snapshots__/types.test.ts.snap
@@ -3,6 +3,10 @@
 exports[`generate types correctly generates JS SDK types with custom data type 1`] = `
 "import { DevCycleJSON } from '@devcycle/js-client-sdk'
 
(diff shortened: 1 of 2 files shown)


src/commands/variations/__snapshots__/create.test.ts.snap
```diff
@@ -74,7 +74,9 @@ exports[`variations create creates a variation and returns the full feature in i
 `;
 
 exports[`variations create errors when missing required flags in headless mode 1`] = `
-"key is a required field
+"(node:26936) Warning: Mismatched version in @devcycle/cli plugin manifest. Expected: 5.19.0 Received: 5.18.0
```
kaushalkapasi, line 77: bug / bad commit? 🤔

ajwootto: dunno this is what happened when I ran tests

PR #445, opened 2025-03-25
chore: update axios dep

jonathannorris:


JamieSinn (approved):
I think there's some changes between 1.6 and 1.8 that might need more edits but lets see?

jonathannorris:
@JamieSinn it was using `1.7.5` previously

PR #452, opened 2025-07-09
feat: added stale comment for stale features in generate types

jsalaber:


src/commands/generate/types.ts
```diff
@@ -430,6 +449,39 @@ function isVariableDeprecated(variable: Variable, features: Feature[]) {
...
+
+function getRecommendedValueForStale(variable: Variable, feature: Feature) {
+    if (!feature) {
+        return variable.defaultValue
+    }
+    const reason = feature.staleness?.reason
+    if (reason === 'unused') {
+        return variable.defaultValue
+    } else if (reason === 'released') {
+        if (feature.staleness?.metaData?.releaseVariation) {
```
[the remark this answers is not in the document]
jsalaber: would maybe like to add that as metadata from the backend instead of figuring out here in the cli

src/commands/generate/types.ts
```diff
@@ -176,11 +179,22 @@ export default class GenerateTypes extends Base {
...
+        const staleFeatures = await fetchFeatures(
             this.authToken,
             this.projectKey,
+            {
+                staleness: 'all',
+            },
         )
 
+        this.features = [
+            ...(await fetchAllCompletedOrArchivedFeatures(
```
suthar26, line 191: promise.all these two so we its not a waterfall

suthar26 (approved):

src/commands/generate/types.ts
```diff
@@ -176,10 +179,18 @@ export default class GenerateTypes extends Base {
             )
         }
 
-        this.features = await fetchAllCompletedOrArchivedFeatures(
-            this.authToken,
-            this.projectKey,
-        )
+        this.features = await Promise.all([
```
suthar26, line 182: nit:
instead of the .then() you can just get the completedFeatures and staleFeatures as a return value and then have the this.features be a spread of both like you had before

jonathannorris (approved):

PR #465, opened 2025-07-18
fix: Zodios Validation Errors in MCP Tools

jonathannorris:
# Fix Zodios Validation Errors in MCP Tools

## Problem

MCP tools were failing with "Zodios: Invalid response from endpoint" errors despite successful HTTP 200 API responses. Schema validation failed due to missing properties, type mismatches, and incomplete comparator enums in Zod schemas.

### Core Changes
- **New `handleZodiosValidationErrors` utility** - Extracts response data from validation errors when HTTP succeeds but schema validation fails
- **Enhanced Zod schemas** - Added missing properties (`_audiences`, `startWith`/`endWith` comparators) 
- **Environment controls** - `ENABLE_OUTPUT_SCHEMAS` and `ENABLE_DVC_MCP_DEBUG` for debugging and Cursor compatibility


src/api/apiClient.ts
```diff
@@ -109,10 +109,18 @@ export const errorMap = (issue: ZodIssueOptionalMessage, ctx: ErrorMapCtx) => {
     }
 }
 
-export const apiClient = createApiClient(BASE_URL, {
+// TLDR: the inferred TS schema was too big, so this is a workaround to fix it.
```
JamieSinn, line 112: so what does this actually mean for the result/effect of this? mostly curious

jonathannorris: basically removes a typescript compiler warning where the infered type schema was too long, this basically sets the type to a defined type.

JamieSinn: so what does a type schema that's too long cause?

JamieSinn: ah got it

PR #481, opened 2025-08-01
Update mcp auth screen

cdruxerman:
- update mcp auth screen to match styles of the auth0 screens

### DIFF
--- a/mcp-worker/src/consentScreen.ts
+++ b/mcp-worker/src/consentScreen.ts
@@ -29,9 +29,21 @@ export function renderConsentScreen({
                     content="width=device-width, initial-scale=1"
                 />
                 <style>
-                    @import url('https://fonts.googleapis.com/css2?family=Nunito:wght@800&display=swap');
+                    @font-face {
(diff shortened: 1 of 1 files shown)


mcp-worker/src/consentScreen.ts
```diff
@@ -21,180 +21,450 @@ export function renderConsentScreen({
...
+        <html lang="en">
             <head>
-                <title>Authorization Request</title>
+                <title>DevCycle - Authorization Request</title>
                 <meta
                     name="viewport"
                     content="width=device-width, initial-scale=1"
                 />
                 <style>
+                    @import url('https://fonts.googleapis.com/css2?family=Nunito:wght@800&display=swap');
```
cdruxerman, line 32: @jonathannorris this is one of the suggested changes from jamie. We were using webflow links for the auth0 pages. I asked cursor to replace them, but it couldn't find uni-neue available. Where do we pull our fonts from?

jonathannorris (approved):

